PR #46, opened 2016-05-14
E-Ballots

BenMusch:
Adds e-balloting

Things I changed in the code:
- Added `ballot_code` column to Judges. Currently, it's generated when Judges are created from the batch import or from the web interface. Maybe I should switch it so that's it's always done on creation for the judges?
- Added an `EBallotForm`, that inherits from `ResultEntryForm` and adds additional validation logic
- Added whitelisted routes `/e_ballots` to search for the current e_ballot, and `/e_ballots/{ballot_code}` to submit an individual ballot
- Made small changes to `round_entry.html` to add some fields & javascript for the e-ballots

Here's an overview of how it works:

Judges are assigned ballot codes upon creation, which can be viewed in the Judges List view by their name:

image

Judges can go to `/e_ballots` to enter their ballot code:

image

Then, the ballot for the current round with that ballot code it brought up (or an error message if the code was incorrect or the judge isn't paired in). The only change to the round entry form is that there are a few extra validations, and that there is a ballot_code box for judges to confirm their ballot code. This is a bit redundant, since the code is in the URL, so it might be worth removing in the future.

image

Hopefully I'll be able to test this out at APDA summer, which is being hosted at NU.

Here's some things I wanna add to this in the next few weeks:
- Tab settings: Add an option to turn off e-ballots
- Speaks justifications: Add an option for the tab director to change the minimum and maximum speaks that need justification. Right now, the value for TabSettings must be an integer. The possible solutions to this are
1. Add a separate min_speaks and max_speaks integer field to the TabSettings model. I think this is preferable, because it won’t interfere with anything else if we set a default value
2. Enter the min and max speaks as an int, so that 23.25 would be represented as 2325. Less preferable in my opinion, just because it would be counter-intuitive for both the tab directors and developers in the future.
3. Change the field to a float, also less preferable in my opinion because it is inaccurate for most settings, and because it could require lots of other changes in the app
- The nav-bar: When the user isn’t logged in, we should just display a link to “pairings” and “e-ballots”, rather than showing all of the options (even though it will still force a login if the user tries to go to another page)
- Judge listings: Add javascript to toggle whether or not ballot codes are shown, to avoid judges stealing others’ codes. This isn’t a huge issue, though, since you can’t re-submit ballots via e-balloting, so if a judge sees that a ballot for their round has been submitted already, they can just go to tab and resolve the issue.


BenMusch:
not sure why travis is failing now, the sqlite db has the column and there's a south migration file for the field

BenMusch:
Sorry for all the commits. . . gonna remove the unnecessary changes then rebase soon

mittab/static/css/bootstrap-grid.css
```diff
@@ -0,0 +1,950 @@
...
+}
+*,
+*:before,
+*:after {
+  -webkit-box-sizing: border-box;
+  -moz-box-sizing: border-box;
+  box-sizing: border-box;
+}
+
+.team_container {
```
BenMusch: This stuff should definitely be in stylesheet.css, but the style wasn't showing up when I put it in there for some reason

mittab/apps/tab/middleware.py
```diff
@@ -1,15 +1,29 @@
...
+
+e_ballot_regex = re.compile("/e_ballots/\S+")
 #backend_white_list = ('/static/css/stylesheet.css', '/enter_tab/', '/static/images/title_banner.png')
 
 class Login:
     "This middleware requires a login for every view"
     def process_request(self, request):
-        if request.path not in login_white_list and request.user.is_anonymous():
+        whitelisted = (request.path in login_white_list)
+        whitelisted = whitelisted or self._is_eballot_url(request.path)
```
jolynch: I think this needs to just be self._is_eballot_url(request)

and then later instead of request == ... do request.path ==

mittab/apps/tab/models.py
```diff
@@ -38,43 +54,65 @@ def delete(self):
...
             raise Exception("Debater on teams: %s" % ([t.name for t in teams]))
 
+
 class Team(models.Model):
-    name = models.CharField(max_length=30, unique = True)
+    """
+    Contains information with regard to the team itself. Contains the team name, the team school, the debaters in that
+    team, the team seed, and the information on whetehr that team is checked in or not.
+    """
+    alphanumeric = RegexValidator(r'^[0-9a-zA-Z ]*$', 'Only alphanumeric characters are allowed.')
```
jolynch: Why do we need this validator? Also we at least need to have "/" in the allowable set since a lot of teams do hybrids as SchoolA/SchoolB

BenMusch: Agreed, I think this should be deleted

mittab/apps/tab/templates/all_tab_cards.html
```diff
@@ -1,13 +1,15 @@
...
 <script type="text/javascript" src="http://ajax.googleapis.com/ajax/libs/jquery/1.6.4/jquery.min.js"></script>
 <script type="text/javascript" src="http://ajax.googleapis.com/ajax/libs/jqueryui/1.8.16/jquery-ui.min.js"></script>
 <script type="text/javascript" src="/static/javascript/tab.js"></script>
 
 <div id="progressbar" class="printable" data-count="0" data-max="{{ all_teams|length }}"></div>
 
 {% for team in all_teams %}
-<div class="printable tab_card_item {% if forloop.counter|divisibleby:2 %}page_break_after{% endif %}" data-team-id={{team.id}}></div>
-
+    <div id="tab-cards" class="printable tab_card_item {% if forloop.counter|divisibleby:2 %}page_break_after{% endif %}">
```
jolynch: This has an extra ">" bracket which prevents the proper rendering of the divs -> prevents javascript from filling in all_tab_cards.

I believe you want to nix the ">" after the endif

mittab/apps/tab/team_views.py
```diff
@@ -43,228 +56,269 @@ def view_team(request, team_id):
...
 @permission_required('tab.tab_settings.can_change', login_url="/403/")
 def all_tab_cards(request):
+    """
+    Displays all the tab cards. And yes, I will editorialise that all tournaments should release tab cards on the ADPA
+    forum to the public. However, it is something which should clearly be limited in the tab settings for sure.
+    """
     all_teams = Team.objects.all()
     return render_to_response('all_tab_cards.html',
-                              locals(),
+                              all_teams,
```
jolynch: This needs to be a dict {'all_teams': all_teams}

mittab/apps/tab/models.py
ifly6: I added this because if you try to use non-ASCII characters, Python complains when it is `str`-ed in `team_views.py`, preventing someone from adding the team.

mittab/apps/tab/templates/all_tab_cards.html
ifly6: I did this by accident. My apologies.

BenMusch:
@ifly6 Just a heads-up, I'm gonna open a new PR without the changes to code style from here. They're definitely helpful, but it makes it harder to sift through and find the changes that I made. I can add them into a separate PR later by cherry-picking the commits

BenMusch:
Closing to re-open on a PR that's easier to sift through

PR #48, opened 2017-03-03
More excel import-export, along with some visual changes

ifly6:
mittab-nav-bar-change

Implemented some import-export stuff from Excel. There is now a page
where you can download dynamically generated excel files that accompany
the rooms, judges, and teams which were imported, along with excel
files containing team and speaker statistics so tournaments which don’t
follow the standard MIT-TAB ranking can get the data they want.

mittab-export-screen
mittab-export-team-stats

In the upload data form, I’ve added text which tells people what the
headers are, so they don’t have to go digging in the manual to find out.

mittab-upload-screen

The import system has also been reworked so that it overwrites
preexisting data rather than simply skipping the entry. This means that
changes applied in some excel document can be applied en-masse by
importing the changed document.

Also, implemented some visual changes—for example, there was an issue
with the upload boxes in the upload screen, the error pages were a bit
off in some places, some of the CSS put some objects in the wrong place
(e.g. small screens, the pair next round button overlapped with another
button). I’ve also changed the background of all pages from sky-blue to
a more neutral grey, but that can easily be reverted.

Addendum: I've also added a change to the pairing error view, so it actually displays the type of error encountered. This should be much more helpful to people who are trying to solve issues.

mittab-new-error-screen

EDIT: Added photos.


ifly6:
I've run tests with the import-export, importing and exporting the various files with various configurations and successfully run a test tournament to populate data for the speaker-teams stat table, which exported successfully as well.

PR #49, opened 2017-04-24
Update fixtures for new speaker scale

BenMusch:
At nats, the speaker scale was changed from 23-> 27 to 5->45. I ran a script to update the fixtures so that data now reflects the new scale


BenMusch:
Closing this to revert changes to sqlite

PR #51, opened 2017-04-24
changed validate_speaks, added ability to set bounds in tab settings

ifly6:
changed the pairing_db_clean database to automatically include that tabsetting in the clean database. the upper_bound var is set to 45 and the lower_bound var is set to 5.

the main changes that it made, which my ide didn't then autoformat :/ , is in the validate speaks section under `forms` and then the `start_new_tourney` method under `pairing_views`


mittab/apps/tab/forms.py
```diff
@@ -39,90 +44,104 @@ def __init__(self, *args, **kwargs):
...
+
 def validate_speaks(value):
-    if not (21.0 <= value <= 29.0 or value == 0):
-        raise ValidationError(u'%s is an entirely invalid speaker score, try again.' % value)
-    
-#TODO: Rewrite this, it is ugly as hell
+    lower_bound = 5   # defaults
+    upper_bound = 45
+    try:
+        lower_bound = int(TabSettings.objects.get(key="lower_speaks").value)
```
BenMusch: Small thing: The `value` field in the `TabSettings` model is already an int, so there's no need to cast it to an int here

BenMusch:
You might wanna run `python manage.py dumpdata mittab/apps/tab/fixtures/testing_db.json` so that the fixtures are updated to include the tabsetting

ifly6:
^ Done

mittab/apps/tab/fixtures/testing_db.json
```diff
@@ -1 +1 @@
...d": 0, "name": "Penn DS", "debaters": [131, 132], "checked_in": true}}, {"pk": 66, "model": "tab.team", "fields": {"school": 15, "seed": 0, "name": "Smith Ravenclaw", "debaters": [128, 129], "checked_in": true}}, {"pk": 67, "model": "tab.team", "fields": {"school": 15, "seed": 1, "name": "Smith Slytherin", "debaters": [133, 134], "checked_in": true}}, {"pk": 68, "model": "tab.team", "fields": {"school": 14, "seed": 0, "name": "Tufts EK", "debaters": [135, 136], "checked_in": true}}, {"pk": 69, "model": "tab.team", "fields": {"school": 23, "seed": 1, "name": "Georgetown Bend and Snap", "debaters": [137, 138], "checked_in": true}}, {"pk": 70, "model": "tab.team", "fields": {"school": 14, "seed": 1, "name": "Tufts ES", "debaters": [139, 142], "checked_in": true}}, {"pk": 71, "model": "tab.team", "fields": {"school": 16, "seed": 1, "name": "W&M Django's Novices", "debaters": [140, 141], "checked_in": true}}, {"pk": 72, "model": "tab.team", "fields": {"school": 14, "seed": 0, "name": "Tufts Bendrew Husick", "debaters": [143, 146], "checked_in": false}}, {"pk": 73, "model": "tab.team", "fields": {"school": 42, "seed": 1, "name": "W&M/UAlbany Hybrid", "debaters": [144, 145], "checked_in":
```
BenMusch: Any idea why this file was cleared? It contains all of the teams used to test the pairing logic, and it seems like it was wiped to a blank slate

mittab/apps/tab/forms.py
```diff
@@ -39,90 +44,104 @@ def __init__(self, *args, **kwargs):
...
+        upper_bound = int(TabSettings.objects.get(key="upper_speaks").value)
+    except Exception as e:
```
BenMusch: Why is this surrounded by a try/except? If an error occurs here, we probably want the transaction to fail.

mittab/apps/tab/forms.py
```diff
@@ -39,90 +44,104 @@ def __init__(self, *args, **kwargs):
...
 def validate_speaks(value):
-    if not (21.0 <= value <= 29.0 or value == 0):
-        raise ValidationError(u'%s is an entirely invalid speaker score, try again.' % value)
-    
-#TODO: Rewrite this, it is ugly as hell
+    lower_bound = 5   # defaults
+    upper_bound = 45
+    try:
+        lower_bound = int(TabSettings.objects.get(key="lower_speaks").value)
+        upper_bound = int(TabSettings.objects.get(key="upper_speaks").value)
```
BenMusch: There's already a method in TabSettings that lets you do this more elegantly:

```python
lower_bound = TabSettings.get('lower_speaks', 0)
upper_bounds = TabSettings.get('upper_speaks', 50)
```

This also lets you specify a default value, which makes the code less error-prone

mittab/libs/tab_logic.py
```diff
@@ -286,6 +286,7 @@ def validate_round_data(round_to_check):
     If any of these fail we raise a specific error as to the type of error
     """
 
+    # TODO (Kevin) remove this? Can just assign judges later.
```
BenMusch: This error is used to update the UI and display a message that there aren't enough judges. Using errors for control flow probably isn't the greatest thing, but that's something to keep in mind before changing this

BenMusch:
I committed `a2ec422` to handle the merge conflict that came from my quick-fix. Shouldnt change anything important

mittab/apps/tab/forms.py
BenMusch: Also, when you use a try-except, make sure to specify the error class. Otherwise something like a typo in a function name could sleep by and we'd never notice

PR #54, opened 2017-04-26
Django 1.11 upgrade + Integration tests to insure functionality

BenMusch:
Ran through the migrations guides. The main contributions in this PR are to get integration testing to make sure no small but important things broke in the upgrade

Important Django changes reflected in the code:

 - URLs must be named to be referenced
 - `render_to_response` was deprecated
 - Django commands option syntax has to be changed, optional arguments are added in the `add_arguments` method, rather than as a class variable
 - Model forms must specify which fields they updated (setting `fields = '__all__'` invokes the traditional behavior)
 - `simplejson` isn't included by default
 - `South` is deprecated, django has migrations natively
 - `ModelMultipleChoiceField` doesn't work in chrome with the Filter widget :( 
 - mimetypes in responses are replaced with subclasses (i.e. `JsonResponse`)
 - `FileWrapper` was moved to `wsgiref.util`
 - `settings.py` formatting for template info changed


BenMusch:
@jolynch Any concerns about deploying if I end up merging this upgrade?

BenMusch:
Closing, will opt for a more incremental approach

PR #56, opened 2017-06-04
Bash script to automate PythonAnywhere server setup

BenMusch:


jolynch:
Not sure how relevant it is, but heads up that you can also run the intialize_tourney command, e.g.

```
python manage.py initialize_tourney ...
```

PR #76, opened 2017-09-10
Fix error in judge import with blank phone

BenMusch:


BenMusch:
Closing, duplicate of https://github.com/jolynch/mit-tab/pull/75

PR #78, opened 2017-09-14
Add speaks distribution script

JEddy92:
I've been manually making the raw speaks data a list of strings, not sure if there's an easier way to do that to align with the data output from mit-tab.


BenMusch:
@JEddy92 Thanks! Will merge & then hopefully get this to work directly from the mit-tab dataset

PR #120, opened 2018-02-05
[WIP] Consistent style across view files

BenMusch:


ifly6:
By the way, isn't it the case that PyCharm and similar IDEs can do all of this formatting automatically?

BenMusch:
@ifly6 Yeah 🤷‍♀️ but I don't feel like downloading pycharm & also prefer it to not be automatic

I'd rather integrate pylint into the travis build so that builds fail with invalid style: https://www.pylint.org/

PR #132, opened 2018-03-09
Emojis!

BenMusch:
The people have spoken


BenMusch:
I believe this fixes https://github.com/jolynch/mit-tab/issues/32

PR #139, opened 2018-04-09
[WIP] Add room groups

BenMusch:
- Categorize rooms with different tags
 - Add debaters/judges who get priority for certain room groups (e.g. tab directors get the rooms near tab, accessibility issues, etc.)
 - Check-out rooms in bulk (useful for rooms you don't have for all rounds)

TODO:
 - Add UI for modifying RoomGroups


BenMusch:
Closing this because I think the design should be different. Should be two models instead of 1:

`RoomTag`: Used for things like acceessibility
`RoomGroup`: Used from check-ins

PR #142, opened 2018-05-30
[WIP] UI Redesign, Bootstrap 4, Django 1.10

BenMusch:
The UI is getting a bit too outdated to easily update

Moreover, the styling is quite large, which makes it pretty hard to add new features since there's not much room on the nav, etc.

Focusing on the layout/structure now. Will finish up with the colors


BenMusch:
Probably gonna leave this alone, but keep the PR around for reference as I change things more incrementally

PR #158, opened 2019-02-26
Python3

BenMusch:
This really has to happen soon...


BenMusch:
Things seem to be working 🍾

Tested the pairings against those generated by the python2 version and they're identical

PR #174, opened 2019-04-11
Feat/user page

yoitsdave:
This is copy-pasted from the commit message:

Introduce a page at /public_status/team_name/<team_name>/ and 
/public_status/debater_name/<debater_name>/ which displays some 
information teams will want about the upcoming round. Uses the same 
check as /pairings/pairinglist/ to ensure pairings are published and 
ought be publicly accessible. To provide user access to these pages, add 
a new Next Round Look-up section to the login page with text-boxes 
linking to lookup by debater name or team name.

### DIFF
--- a/mittab/apps/tab/team_views.py
+++ b/mittab/apps/tab/team_views.py
@@ -380,20 +380,15 @@ def team_stats(request, team_id):
     return JsonResponse(data)
 
 
-def public_status_by_team(request, team_name, team_ref=None):
-    if team_ref:
-        matches = [team_ref]
-    else:
-        clean = team_name.replace('_', ' ')
-        matches = [*Team.objects.filter(name__iexact=clean)]
+def public_status(request, team_id, team_ref=None):
+    matches = Team.objects.filter(pk=team_id)
     
     if matches:
         round_number = TabSettings.get('cur_round') - 1        
         pairing_exists = TabSettings.get('pairing_released', 0) == 1
 
         if pairing_exists:
-            return _public_status_paring_exists(request, team_name, matches, 
-                                                round_number)
+            return _public_status_pairing_exists(request, matches, round_number)
         else:
             out = {
                     'no_navigation': True,
@@ -417,9 +412,7 @@ def public_status_by_team(request, team_name, team_ref=None):
                                   context_instance=RequestContext(request))
                                   
 
-def _public_status_paring_exists(request, team_name, matches, round_number):
-    # Note here that the queries returned by Model.objects.filter are evaluated
-    # lazily, so there is no performance penalty to defining them all up here
+def _public_status_pairing_exists(request, matches, round_number):
     gov_pairings = Round.objects.filter(round_number=round_number, 
                                         gov_team=matches[0])
     opp_pairings = Round.objects.filter(round_number=round_number, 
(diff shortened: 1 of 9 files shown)


mittab/apps/tab/team_views.py
```diff
@@ -380,3 +380,120 @@ def team_stats(request, team_id):
...
+    # Note here that the queries returned by Model.objects.filter are evaluated
+    # lazily, so there is no performance penalty to defining them all up here
+    gov_pairings = Round.objects.filter(round_number=round_number, 
+                                        gov_team=matches[0])
+    opp_pairings = Round.objects.filter(round_number=round_number, 
+                                        opp_team=matches[0])
+    byes = Bye.objects.filter(round_number=round_number,
+                              bye_team=matches[0])   
+        
+    if gov_pairings:
```
BenMusch: perhaps a nitpick, but I'd prefer to move this logic to the template rather than in the view, since a lot of it feels a lot more like "how to I display this thing" sorts of issues. So the logic in the team_views.py would roughly be:

```
if pairing is a round:
	render('public_status', is_round=True, obj=round)
elif pairing is a bye:
	render('public_status', is_round=False, obj=bye)
else:
	render error
```

Then in the view we have an if statements to do the rest of the stuff and keep the `team_views.py` file relatively slim

mittab/apps/tab/team_views.py
```diff
@@ -380,3 +380,120 @@ def team_stats(request, team_id):
...
+                                  context_instance=RequestContext(request))
+
+    return render_to_response('public_status.html', out,
+                          context_instance=RequestContext(request))
+
+def _judge_names(judges):
+    return ', '.join(judge.name for judge in judges.all())
+    
+def public_status_by_debater(request, debater_name):
+    matches = [*Debater.objects.filter(name__iexact=debater_name.replace('_', ' '))]
```
BenMusch: 1. Same comment here about replacing names
 2. Filter results are already iterable, what's the reason for converting these to arrays?

mittab/apps/tab/team_views.py
```diff
@@ -380,3 +380,120 @@ def team_stats(request, team_id):
...
+    else:
+        return render_to_response('error.html',
+                                 {'error_type': 'Tab Card','error_name': 'View',
+                                  'error_info': "That's not a registered team!",
+                                  'no_navigation': True},
+                                  context_instance=RequestContext(request))
+                                  
+
+def _public_status_paring_exists(request, team_name, matches, round_number):
+    # Note here that the queries returned by Model.objects.filter are evaluated
```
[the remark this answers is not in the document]
BenMusch: Oh also, before this gets finalized, can you remove the code comment? Just so there aren't comments hanging around that aren't consistently relevant to the code

yoitsdave:
I wrote the searchable select field assuming that its ok to leak the primary keys of each debater, as the order they were put in shouldn't really affect anything - if that's wrong, let me know and it's easy to switch.

BenMusch:
Thanks! I'll get back to you in like ~1-1.5 weeks, I'm in the middle of finals rn so I'm pretty busy

yoitsdave:
I'm putting this here only because it seems relevant to a focus on improving debater experience rather than tab room experience of mit-tab, but it may deserve its own pull request. 

Considering the discussion currently happening on the forums with regard to tournaments regularly failing to post tab cards on a timely basis [(see here)](http://apda.online/forum/topic/closed-speaks/?part=1#postid-3872), is there any reason not to put `/all_tab_cards/` into the not-logged-in whitelist after all rounds have been finished?

BenMusch:
@yoitsdave this is one of the situations where it's more of a league-wide discussion than a technical one -- I don't really have a strong opinion on it but I could see some people perhaps not wanting to have to judge a team that knows you gave them low speaks in outrounds, for example. Also -- change should probably move to a separate PR so we don't bundle up a bunch of features into this one, making it harder to merge without any issues

BenMusch:
@yoitsdave Actually, gonna have to just say no to the public tab cards things (for now), because of tournaments like nats where the break is secret until after the rounds have completed. In theory, we can add a TabSetting like `dont_release_tab_cards` or something, but in practice, someone would forget to set it and it'd lead to leaking a break that wasn't meant to be leaked

BenMusch:
That said, I'm totally in favor of a more complex flow that would work something like displaying a "release tab cards" button when all rounds are done

mittab/apps/tab/team_views.py
```diff
@@ -380,3 +380,120 @@ def team_stats(request, team_id):
...
+    # lazily, so there is no performance penalty to defining them all up here
+    gov_pairings = Round.objects.filter(round_number=round_number, 
+                                        gov_team=matches[0])
```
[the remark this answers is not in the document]
yoitsdave: This should be addressed in the latest commit by the move to urls by team primary key - `len(matches) in [0, 1]` will always be true.

mittab/apps/tab/team_views.py
yoitsdave: I've moved a lot of the logic to template, but not quite all of it. In particular, it is much easier to process the judge list into a string in `team_views.py`, so I've kept that in there.

PR #176, opened 2019-04-28
Individual Round Info Pages For Each Team

yoitsdave:
I accidentally pulled some git fuckery with my fork, so I have to reopen this PR :(. Sorry for the clutter

PR #206, opened 2019-05-13
Round info pages for each team (based on #176)

BenMusch:
Updates #176 to work with the new UI

cc @yoitsdave


BenMusch:
Closing... need to rebase instead of merging with master to things get to a more sane state

PR #209, opened 2019-05-13
Reformat mittab/libs/*, add python linter

BenMusch:


BenMusch:
This is too much for me to be confident in it w/o a larger test suite. Gonna add linting more incrementally in another PR

PR #226, opened 2019-06-10
Add ability to export as a Tournament Archive

tienne-B:
The Tournament Archive or DebateXML is an XML-based format for the interchange of debate tournament data. Its schema is currently under review in TabbycatDebate/dta-spec#4.

This commit adds an admin form to export the site's data through that specification for archival and analysis. The page is available through `/archive/download/`, which asks for some supplementary details, such as the tournament's name.

A link to the form should be in the navigation; however, it doesn't seem to be there.

Byes are not yet handled.

### DIFF
--- a/mittab/apps/tab/archive.py
+++ b/mittab/apps/tab/archive.py
@@ -1,0 +1,150 @@
+from xml.etree.ElementTree import Element, SubElement, tostring
+
+from mittab.apps.tab.models import School, Team, Judge, Room, Round
+
+DEBATE_ID_PREFIX = "D"
+ROOM_ID_PREFIX = "V"
+SPEAKER_ID_PREFIX = "S"
+TEAM_ID_PREFIX = "T"
+JUDGE_ID_PREFIX = "A"
+SCHOOL_ID_PREFIX = "I"
+SPEAKER_STATUS_PREFIX = "SC"
+
+
+class ArchiveExporter:
+
+    def __init__(self, name):
+        self.name = name
+        self.root = None
+
+    def export_tournament(self):
+        self.root = Element("tournament", {"name": self.name})
+
+        self.add_rounds()
+        self.add_participants()
+        self.add_schools()
+        self.add_rooms()
+        self.add_categories()
+
+        return tostring(self.root)
+
+    def add_rounds(self):
+        pos = ["pm", "lo", "mg", "mo"]
+        qs = Round.objects.all().order_by("round_number").prefetch_related(
+            "judges", "roundstats_set")
+
+        cur_round = None
+        r_tag = None
+        for debate in qs:
+            if debate.round_number != cur_round:
+                r_tag = SubElement(self.root, "round", {
+                    "name": "Round %s" % debate.round_number
+                })
+                cur_round = debate.round_number
+
+            adjs = " ".join(
+                [JUDGE_ID_PREFIX + str(j.id) for j in debate.judges.all()])
+            d_tag = SubElement(r_tag, "debate", {
+                "id": DEBATE_ID_PREFIX + str(debate.id),
+                "chair": JUDGE_ID_PREFIX + str(debate.chair_id),
+                "adjudicators": adjs,
+                "venue": ROOM_ID_PREFIX + str(debate.room_id)
+            })
+
+            gov_tag = SubElement(d_tag, "side", {
+                "team": TEAM_ID_PREFIX + str(debate.gov_team_id)
+            })
+            opp_tag = SubElement(d_tag, "side", {
+                "team": TEAM_ID_PREFIX + str(debate.opp_team_id)
+            })
+
+            if debate.victor == Round.GOV_VIA_FORFEIT or \
+                    debate.victor == Round.ALL_DROP:
+                opp_tag.set("forfeit", "True")
+            elif debate.victor == Round.OPP_VIA_FORFEIT or \
+                    debate.victor == Round.ALL_DROP:
+                gov_tag.set("forfeit", "True")
+
+            team_points = [0, 0] # Gov, Opp
+            stats = sorted(list(debate.roundstats_set.all().values()),
+                           key=lambda x: pos.index(x["debater_role"]))
+
+            if debate.victor == Round.GOV or debate.victor == Round.OPP:
+                for i, speech in enumerate(stats):
+                    side_tag = gov_tag if i % 2 == 0 else opp_tag
+                    team_points[i % 2] += speech["speaks"]
+                    speech_tag = SubElement(side_tag, "speech", {
+                        "speaker": SPEAKER_ID_PREFIX + str(speech["debater_id"]),
+                        "reply": "False"
+                    })
+                    ballot = SubElement(speech_tag, "ballot", {
+                        "adjudicators": adjs,
+                        "rank": str(speech["ranks"]),
+                        "ignored": "False"
+                    })
+                    ballot.text = str(speech["speaks"])
+
+            gov_win = [Round.GOV, Round.GOV_VIA_FORFEIT, Round.ALL_WIN]
+            gov_rank = "1" if debate.victor in gov_win else "2"
+            gov_ballot = SubElement(gov_tag, "ballot", {
+                "adjudicators": adjs, "rank": gov_rank})
+            gov_ballot.text = str(team_points[0])
+
+            opp_win = [Round.OPP, Round.OPP_VIA_FORFEIT, Round.ALL_WIN]
+            opp_rank = "1" if debate.victor in opp_win else "2"
+            opp_ballot = SubElement(opp_tag, "ballot", {
+                "adjudicators": adjs, "rank": opp_rank})
+            opp_ballot.text = str(team_points[1])
+
+    def add_participants(self):
+        participants_tag = SubElement(self.root, "participants")
+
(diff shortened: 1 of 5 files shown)


mittab/apps/tab/archive.py
```diff
@@ -0,0 +1,153 @@
+from xml.etree.ElementTree import Element, SubElement, tostring
+
+from .models import School, Team, Judge, Room, Round
```
BenMusch:
```suggestion
from mittab.apps.tab.models import School, Team, Judge, Room, Round
```

Nitpick, just for the sake of consistency with other files

mittab/apps/tab/archive.py
```diff
@@ -0,0 +1,153 @@
...
+TEAM_ID_PREFIX = "T"
+JUDGE_ID_PREFIX = "A"
+SCHOOL_ID_PREFIX = "I"
+SPEAKER_STATUS_PREFIX = "SC"
+
+
+class ArchiveExporter:
+
+    def __init__(self, request, form):
+        self.request = request
```
BenMusch: Looks like the request is unused? Maybe we can take that out of this class

BenMusch: Actually, more generally, what do you think about refactoring this just to take a `host`, `name` and `start` parameter, and ditch the dependency on the form and request entirely?

mittab/apps/tab/forms.py
```diff
@@ -474,3 +474,9 @@ def score_panel(result, discard_minority):
     pprint.pprint(ranked)
 
     return ranked, final_winner
+
+
+class ArchiveDownloadForm(forms.Form):
+    name = forms.CharField(max_length=100, required=True)
```
BenMusch: Can we restrict this to alpha-numeric values with no spaces? Since it goes directly into an http header I'd like it to be super sanitized.

BenMusch:
Here is how the back-ups happen right now: https://github.com/MIT-Tab/mittab-deploy/blob/master/bin/backup

Just a script that downloads the database file and uploads it to DigitalOcean's storage. Hoping to automate things and make them a bit more robust this summer.

mittab/apps/tab/archive.py
```diff
@@ -0,0 +1,153 @@
...
+            elif debate.victor == Round.OPP_VIA_FORFEIT or \
+                    debate.victor == Round.ALL_DROP:
+                gov_tag.set("forfeit", "True")
+
+            team_points = [0, 0] # Gov, Opp
+            stats = sorted(list(debate.roundstats_set.all().values()),
+                           key=lambda x: pos.index(x["debater_role"]))
+            for i, speech in enumerate(stats):
+                side_tag = gov_tag if i % 2 == 0 else opp_tag
+                team_points[i % 2] += speech["speaks"]
```
BenMusch: Not sure what the expected behavior is here, but this may not behave properly in the case of iron-people. This code is pretty messy, but you may want to look in here for some context on how scoring is done: https://github.com/MIT-Tab/mit-tab/blob/master/mittab/libs/tab_logic/stats.py

This code will work 99% of the time, but I'm not certain about 100%. Might be worth refactoring the tab_logic.stats interface to be something more useful, and then using those functions here. Specifically, functions with these signatures would be great:

```
def debater_speaks_for_round(debater, round):
	pass

def team_speaks_for_round(team, round):
	pass
```

tienne-B: This is to create a "team score" by summing the speaks of the members in the round.

From what I saw, iron speakers have `RoundStats` objects for each of their positions, which is then averaged for the speaker standings. These two objects may have different values for speaks, of which should be captured in the total team speaks here. As there would still be `RoundStats` objects for each position in the case of iron-people, this would work.

A `debater_speaks_for_round(debater, round_number)` function would be inappropriate as a consequence, as that couldn't correctly handle iron-people.

Am I mistaken?

tienne-B:
I am not familiar with the archiving tool; more details would be appreciated. Is archiving automatic with tab directors having access to their archives, and when is it triggered? What would the differences between the two approaches, integrating with the archive tool or having a download view, entail?

The aims of having an interchange format like this have been eloquently laid out in [a 2015 article](https://github.com/TabbycatDebate/dta-spec/blob/master/The-Start-of-History.pdf). I think both uses you've said, analysis and migration, are suitable use-cases for this feature/format. However, backups should still be separate with its specificity to the schema used in the app.

BenMusch:
@tienne-B the back-ups are not accessible by tab directors, it's basically a script I run to delete the server which then uploads the sqlite DB to DigitalOcean's version of Amazon S3. The database files are private (only I and APDA's board have access to it), but we could theoretically upload the XML file to a public location

tienne-B:
The XML file could be available without a form; the host and date attributes are optional, and the tournament name could be provided from the sub-domain (just remembered that was why I included `request` in the class).

I don't have much of an opinion between the files being provided globally (through `nu-tab`) or locally (by mit-tab instance) if tournament officials have easy access to them.

BenMusch:
One key difference between mit-tab and tabbycat is that mit-tab doesn't run on heroku, so servers are not free, and instances do not live for more than ~1 month usually. So I think the best option here is to do both of these. Tournament directors should be able to download XML, and it should also be archived for other people to be able to access it. If you make the tournament director endpoint, I will handle the archiving stuff because it's more tightly-coupled with the deployer logic.

Will review this stuff more thoroughly in the upcoming week but the approach you have here is definitely functional enough to be merged once all the work is done

tienne-B:
It's ready!

mittab/apps/tab/archive.py
BenMusch: I think it would be able to handle iron-men, but the tab logic code is all very messy. Let's leave it like this for now, and I'll dig into potentially using tab_logic some other time

mittab/apps/tab/archive.py
```diff
@@ -0,0 +1,148 @@
...
+                    debate.victor == Round.ALL_DROP:
+                opp_tag.set("forfeit", "True")
+            elif debate.victor == Round.OPP_VIA_FORFEIT or \
+                    debate.victor == Round.ALL_DROP:
+                gov_tag.set("forfeit", "True")
+
+            team_points = [0, 0] # Gov, Opp
+            stats = sorted(list(debate.roundstats_set.all().values()),
+                           key=lambda x: pos.index(x["debater_role"]))
+            for i, speech in enumerate(stats):
```
BenMusch: I think we may only want to do this in the case where the round result is `GOV_WIN` or `OPP_WIN` -- there's no application logic which prevents speaks from being associated with `ALL_WIN` rounds, yet those speaks are ignored (and auto-calculated) for tabulation

mittab/apps/tab/archive.py
```diff
@@ -0,0 +1,148 @@
...
+                    "categories": SPEAKER_STATUS_PREFIX + str(debater.novice_status),
+                    "institutions": institutions
+                })
+                debater_tag.text = debater.name
+
+        for judge in Judge.objects.all().prefetch_related("schools"):
+            SubElement(participants_tag, "adjudicator", {
+                "id": JUDGE_ID_PREFIX + str(judge.id),
+                "name": judge.name,
+                "score": str(judge.rank),
```
BenMusch: So, on APDA, judge ranks aren't frequently re-calculated. They're usually just assigned once based on experience and a calibration ballot. I think most tab directors would be uncomfortable with the judge ranks they assign being released in any sort of public format. Would it be possible to leave this out?

mittab/templates/base/_navigation.html
```diff
@@ -71,6 +72,7 @@
             <a class="dropdown-item {%active request view_backups %}" href="{{ view_backups }}">View Backups</a>
             <a class="dropdown-item {%active request backup %}" href="{{ backup }}">Backup Current</a>
             <a class="dropdown-item {%active request tab_cards %}" href="{{ tab_cards }}">View Tab Cards</a>
+            <a class="dropdown-item {%active request download_archive %}" href="{{ download_archive }}">Create DebateXML</a>
```
BenMusch: Can you add alt help here explaining what this is? Something like "Debate XML is a standardized format used across multiple debate events"

mittab/apps/tab/archive.py
BenMusch: Or maybe I can just filter out this data before it's automatically released anywhere public?

BenMusch:
Just a couple comments and then it's good-to-go. Thanks for doing this!

mittab/templates/base/_navigation.html
tienne-B: Done. Wondering why there's a separate `navigation.html` with the same links. Used anywhere?

mittab/apps/tab/archive.py
tienne-B: Thought the judge ranks would be useful for the tab directors to keep in their records, and that's why I added it. It's true however that public disclosure may not always be wanted. It would be easy to filter out after-generation as it is the only attribute called `score`.

mittab/templates/base/_navigation.html
BenMusch: No, I probably just forgot to delete that when I updated the UI. You can leave it as-is

BenMusch:
LGTM 👍 Will merge this soon, thanks a bunch!

PR #245, opened 2019-10-07
Logout button

roddajohn:
Ran into this problem at Columbia — had temporarily logged in to tab on someone's computer and there was no logout button (without going to admin)!

Edited master template, as well as disambiguated two url routes to allow url helper to function properly.

Simply a time saver.

### DIFF
--- a/mittab/templates/base/_navigation.html
+++ b/mittab/templates/base/_navigation.html
@@ -1,6 +1,7 @@
 {% load tags %}
 
 {% url "index" as home %}
+{% url "logout" as logout %}
 
 {% url "enter_school" as enter_school %}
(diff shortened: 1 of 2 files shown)


mittab/templates/base/_navigation.html
```diff
@@ -97,6 +98,9 @@
             <a class="dropdown-item" href="/admin">Admin Interface</a>
           </div>
         </li>
+	<li class="nav-item">
+            <a class="nav-link" href="{{ logout }}">Logout</a>
```
BenMusch: Nit: change indentation for consistency

```suggestion
        <a class="nav-link" href="{{ logout }}">Logout</a>
```

Issue #53, opened 2017-04-25
Add publicly available judge and team lists


ifly6:
In Tabroom, there are publicly available lists to see the judges and the teams at the tournament. It would be fantastic to see a similar table here. It could probably be extended from the current pairings table, with some small buttons under the big button for pairings to display them.

Certainly, it would certainly be better than trying to find judge lists and what-not by going through random Facebook pages.

Perhaps a table on the lines of (here applicable for Nats), showing if they are checked in for some round, the affiliations they have, and their name.

```
Name      | Affiliations                  | R1 | R2 | R3 | R4 | R5 | R6
----------|-------------------------------|----|----|----|----|----|---
Will      | Swat                          | ✓  | ✓  | ✓  | ✓  | ✓  | ✓
David     | GW                            | ✓  | ✓  | ✓  | ✓  | ✓  | ✓
Zoffer    | Harvard                       | ✓  | ✓  | ✓  | ✓  | ✓  | ✓
Diana     | Yale                          | ✓  | ✓  | ✓  | ✓  | ✓  | ✓
Joe       | Columbia                      | ✓  | ✓  | ✓  | ✓  | ✓  | ✓
Sinan     | Princeton                     | ✓  | ✓  | ✓  | ✓  | ✓  | ✓
Mechanick | Brown                         | ✓  | ✓  | ✓  | ✓  | ✓  | ✓
```
For teams, it would be much easier, something like this would be fine. Also, it would help with people finding out exactly who it is they are hitting. It would also be great if a `TabSettings` could be used as a 1-0 flag to determine whether names should be displayed. Would look something like this:

```
Team         | Members                 | Checked In
-------------|-------------------------|-----------
College Team | John Smith, Jane Doe    | ✓
College Team | John Smith, Jane Doe    | ✓
College Team | John Smith, Jane Doe    | ✓
```
Of course, a change could be implemented to simply remove non-checked in teams from the list after they pass the lenient late limit.

PR #246, opened 2019-10-08
Public team judge lists

roddajohn:
Completes #53 

Doesn't do the lenient late addition, can do later when I have time.


BenMusch:
Just tested it out, one suggested change: Can you try adding the auto-scroll option for the view like there is on pairings? So this can be put up in GA during check-in.

roddajohn:
Just added that — as well as stuck the scrolling code in a separate file, and had both the public pairing view, as well as the team and judge view reference that.

PR #247, opened 2019-10-09
Basic team code functionality

roddajohn:
Solves #138  — doesn't provide a mechanism to get the team codes en masse — but this could perhaps be a new issue (same issue with judge codes).

Also changed the judge ballot code code to make it consistent (added a helper function for the settings of a unique ballot code / team code).


mittab/apps/tab/models.py
```diff
@@ -120,6 +120,37 @@ class Team(ModelWithTiebreaker):
     )
     seed = models.IntegerField(choices=SEED_CHOICES)
     checked_in = models.BooleanField(default=True)
+    team_code = models.CharField(max_length=256,
+                                 blank=True,
+                                 null=True,
+                                 unique=True)
+
+    def set_unique_team_code(self):
```
BenMusch: Only thing I'd change, to make this a bit more readable, how about we split the string on `-` and capitalize it

e.g. `"random-word"` becomes `"Random Word"` so it looks more team-name-like

roddajohn: Done!

BenMusch: Haven't run the code, but I think there's two issues with this version:

1. `upper()` will upcase the whole string, resulting in `"RANDOM WORD"` rather than `"Random Word"`
2. This only hits the processing code in the while loop if there's a team with the same name,

maybe a nested function would be a quick-and-clean solution here, ex:

```python
def set_unique_team_code(self):
    def get_processed_name():
          prefix, suffix = haikunator.haikunate(token_length=0).split("-")
          return "%s %s".format(prefix.capitalize(), suffix.capitalize())
 ```

And then you can just change every assignment to `code` as `code = get_processed_name()`

(Lemme know if I messed up my logic here, not actually running the code so I could be messing up)

roddajohn: Oh you're right, whoops, I'll update it shortly.

roddajohn: @BenMusch should be good now!

Issue #248, opened 2019-10-09
Better interface for TabSettings


BenMusch:
Right now tab settings are kinda a bad UX because it requires people use the documentation and read things carefully, have no typos, etc.

I think we can improve this by having a `/settings` page where we have a better UI controlled by a YAML file. For example

```yaml
  judges_public:
    type: boolean
    description: Is the judge list publicly visible?
    default: 0
 hit_team_before:
   type: integer
   description: Penalty in the pairing algorithm for hitting a team you have already hit. Lower values are a stronger penalty
    default: -100000
```

You can imagine a view which reads this file and renders a form based on this configuration

PR #249, opened 2019-10-10
Settings view

roddajohn:
Addresses #248 .  There's some more space on the right side of the form if there's more documentation / information to tell users about.


settings.yaml
```diff
@@ -0,0 +1,20 @@
+- public_teams:
+  name: public_teams
+  description: "1 if the teams list should be public, 0 if not"
+  value: 0
+- public_judges:
+  name: public_judges
+  description: "1 if the judges list should be public, 0 if not"
```
BenMusch: Quick comment, haven't fully reviewed this yet: I think it would be ideal to have a `type` field for cases like this that we set to `boolean`, and then we can render it as a checkbox in the form rather than a number input. I think that would be a bit more of an intuitive UX

If you wanna do it this way as a first iteration and then you (or I) can add the boolean feature on-top of it that's also totally fine

roddajohn: I added checkboxes — lmk what you think!

Issue #218, opened 2019-05-20
Don't allow debaters to be on multiple teams


BenMusch:
Can be enforced by a validation. Leads to bad stat calculations when this happens

PR #250, opened 2019-10-10
Issue 218

roddajohn:
This solves #218

### DIFF
--- a/mittab/apps/tab/forms.py
+++ b/mittab/apps/tab/forms.py
@@ -97,6 +97,11 @@ class TeamForm(forms.ModelForm):
         data = self.cleaned_data["debaters"]
         if len(data) not in [1, 2]:
             raise forms.ValidationError("You must select 1 or 2 debaters!")
+        for debater in data:
+            if debater.team_set.count() > 0:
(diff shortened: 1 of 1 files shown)


mittab/apps/tab/forms.py
```diff
@@ -97,6 +97,11 @@ def clean_debaters(self):
         data = self.cleaned_data["debaters"]
         if len(data) not in [1, 2]:
             raise forms.ValidationError("You must select 1 or 2 debaters!")
+        for debater in data:
+            if debater.team_set.count() > 0:
+                raise forms.ValidationError(
+                    "A debater cannot already be on a different team!"
```
BenMusch: The most common time where this happens is when a team is created, they cancel, and a new team is added with 1 of the debaters of the previous team

To make this message a bit clearer for that use-case, I'd add this to the message:

``"A debater cannot already be on multiple teams! Consider editing the debaters' existing team, or removing them from it before creating this one"

roddajohn: I also think there's a bug here re saving a team, I'll look into it and fix.

PR #254, opened 2019-10-10
Add note in docs that team codes aren't a released feature yet

BenMusch:
cc @roddajohn, feel free to remove this once everything is done


roddajohn:
#247 is merged right ?

BenMusch:
@roddajohn Oh I'm stupid -- I totally missed the html changes at the bottom. I think there's still some more (small) changes to make. Specifically: Team codes should also be added to the public teams lists _and_ to the team drop-down in the ballots form, so that basically no publicly-visible page has any normal team name on it

roddajohn:
Got it — I'll make those changes soon!

roddajohn:
I'm working on it and came across this interesting question — on the public teams page, what do we think about redacting school names (and maybe debater names) if team codes are in use?  I'm thinking right now to add two flags to allow the user to control that.

BenMusch:
@roddajohn IMO, we can show the normal team names and schools there. What matters is for submitting ballots and viewing pairings, where it affects the judge

BenMusch:
Actually, changed my mind. Hide the schools on that page so judges can't look it up if they get curious

roddajohn:
Ha!  We had the same thoughts, I have a toggle that the tournament can set

roddajohn:
See #257 which I think solves this

BenMusch:
Will review this weekend!

PR #257, opened 2019-10-10
Better team codes

roddajohn:
Picking up from #254 

Thinking of also toggling a setting to allow for the use of these team codes on all the backend views — maybe not a good idea to use mismatched names on the front end and the backend.

Probably will just add a setting for it, have it default to no, but allow people to use it if they want.

Haven't checked `pylint` so will do that soon on this PR.  Don't merge yet.


roddajohn:
Ok it's good now.

`pylint` is good, as well as added backend team code functionality with a new flag, and added flags for what's visible on the public team list.

roddajohn:
Probably should've been multiple PRs, but I'm being lazy...

mittab/apps/tab/migrations/0011_auto_20191010_2104.py
```diff
@@ -0,0 +1,17 @@
...
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('tab', '0010_auto_20191008_2358'),
+    ]
+
+    operations = [
```
BenMusch: Is this a git issue? Any reason this change is on this PR?

docs/Advanced-Topics.md
```diff
@@ -62,12 +62,18 @@ you created it
...
 | `judges_public`     | `0`           | `1` if the judges list is publicly visible, `0` if not.                                                                                                                                                                                                                                                          |
 +---------------------+---------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| `schools_public`    | `1`           | `1` if the schools are publicly visible on the public teams list.                                                                                                                                                                                                                                                |
```
BenMusch: Why not get rid of this and use the `use_team_codes` setting to control this? I can't think of a circumstance where people would want to hide schools other than using team codes

docs/Advanced-Topics.md
```diff
@@ -62,12 +62,18 @@ you created it
...
+| `debaters_public`   | `1`           | `1` if the debaters are publicly visible on the public teams list.                                                                                                                                                                                                                                                |
++---------------------+---------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| `team_codes_backend`| `0`           | `1` if the team codes are used for all backend views, `0` if not.                                                                                                                                                                                                                                                |
```
BenMusch: This phrasing probably isn't something most tab directors would understand. Maybe change to (`if team codes are used in non-public views, in addition to the public ones`)

mittab/apps/tab/migrations/0011_auto_20191010_2104.py
roddajohn: Fixed!

docs/Advanced-Topics.md
roddajohn: Yep, agree, check it out now.

docs/Advanced-Topics.md
roddajohn: Check it out now, I changed backedn to non-public.  I didn't add the in addition clause as technically (?) one could only use team codes on the backend using these flags (not that they'd want to).

mittab/apps/tab/migrations/0011_auto_20191010_2104.py
BenMusch: Ah I see -- is the issue that the migration wasn't added in the not-on-multiple-teams PR, but django application logic still checks for uniqueness without the migration, so that PR worked even though the migration wasn't added? And now when you try to run `migrate` it asked you to `makemigrations` first?

BenMusch: ^ If so, just open another PR to add this migration

BenMusch:
LGTM, just two remaining fixes:

1. The team list page hasn't been updated to hide schools after the tab setting logic was changed

2. The docs build was failing on an earlier commit because the table isn't correctly formatted. Make sure everything lines up correctly, and that there are no tab characters where there should be spaces. To test the docs locally:

```
cd docs
pip install -r requirements.txt
make html
```

roddajohn:
Not sure what you mean re the team list page, it's calling the school's display function which reads the Tab setting.

I think I fixed the docs thing.

mittab/apps/tab/migrations/0011_auto_20191010_2104.py
roddajohn: I'll check this out.

roddajohn: #261 solves

Issue #244, opened 2019-10-04
Make it easier to see checked-in teams


yashaslokesh:
**Feature Request**

Templating Cleanup

The route at `/pairings/status` shows a long list of teams that aren't checked in or don't have a bye. I think the interface here could be improved to show a more-condensed view of the teams, so that all the teams that aren't checked in can be seen on one screen.

I have experience with Flask & Jinja2, I could learn Django and update the relevant template.

BenMusch:
Thanks for reporting this!

Out of curiosity, are you familiar with the filers available on the team list view? There's a checkbox that allows you to filter by check-in. I think that may serve this purpose pretty well

yashaslokesh:
I do know about those filters, and they are useful. Specifically on `/pairings/status`, I think there could be a better way to tell Tab about the non-checked in teams, though.

yashaslokesh:
Additionally, we had a problem this past weekend with fake debaters, ones that were not part of any team but were still on the debaters list (input errors). We clicked on each debater and checked if they were part of a team and removed the 'fake' debaters. 

Maybe a notice could be added on the `pairings/status` page or on the `/view_debaters` page indicating all team-less debaters?

BenMusch:
Sounds good. Does this seem like a reasonable description for how the fix should be implemented?

```
On the pairing status page, rather than having one error for each un-paired-in team, there would be one alert total, which can be expanded/collapsed to show/hide the list of all missing teams
```

PR #258, opened 2019-10-11
Fixes 244

roddajohn:
I believe this solves #244 

Two changes:

1) Adds a collapsible alert when there are teams that are checked in but aren't paired in and don't have a bye, in addition to the error containing a count.

2) On the debater list page, and the main page, a debater not on a team will have (NO TEAM) appended to the name, this is searchable on both views.

### DIFF
--- a/mittab/templates/pairing/pairing_control.html
+++ b/mittab/templates/pairing/pairing_control.html
@@ -9,7 +9,7 @@
       <div class="col">
         <h3>
           Round Status for Round {{round_number}}:
-          <small class="{% if not errors %}text-success{% else %}text-danger{% endif %}">
+          <small class="{% if not errors or num_excluded == 0 %}text-success{% else %}text-danger{% endif %}">
             {% if not errors or num_excluded == 0 %} Valid {% else %} Invalid {% endif %} pairing
           </small>
         </h3>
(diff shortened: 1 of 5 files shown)


mittab/templates/pairing/pairing_control.html
```diff
@@ -71,6 +71,15 @@ <h6>Display Options</h6>
           <div class="alert alert-danger">{{ error }}</div>
           {% endfor %}
         {% endif %}
+	{% if num_excluded > 0 %}
```
BenMusch: Small thing: If you look above this, there's some other logic that uses the truthy-ness of `errors` to display `valid` or `invalid`. Can you modify that to also check `num_excluded > 0`

roddajohn: Just committed a change that should resolve this!

mittab/templates/pairing/pairing_control.html
```diff
@@ -10,7 +10,7 @@
         <h3>
           Round Status for Round {{round_number}}:
           <small class="{% if not errors %}text-success{% else %}text-danger{% endif %}">
```
BenMusch: @roddajohn one more spot on this line

roddajohn: Should be good now!

Issue #156, opened 2019-01-17
Cache debater and team rankings


BenMusch:
that page load is a long boi

ifly6:
There is some kind of caching library or something of that sort, isn't there? Is it broken?

BenMusch:
There is, but I haven't looked into the capability it has to store potentially really large lists

Additionaally, the current cache logic is just "cache the result for x seconds", but we may want more advanced logic along the lines of "cache the result until a round has been saved/updated/deleted"

ifly6:
Also, having worked with Django in something similar, it seems to me that this could simply be an optimisation problem in the `tab_logic` file. On cursory glance, many of the operations have a lot more IO overhead than are necessary for their specific calculations.

BenMusch:
@ifly6 Yeah, I haven't really analyzed it too much but I'd imagine that there are loads of places which do a linear number of DB reads/writes which could be constant. For the most part, I'm not _too_ concerned about that, just the tab_logic library is already super hard to read so I'm fine with sacrificing performance for the readability there

ifly6:
Looking at the rank teams response function, for example, it calls a method which calls total wins, speaks, and ranks. And then it discards those results after sorting, before calling those exact same outputs again to produce the final data set. I'd imagine the caching system would produce some efficiencies, but I've never noticed it speeding up on a second reload.

BenMusch:
The caching is super short. It saves data basically long enough for 1 pairing process but doesn't hold onto data for more than a few seconds. So the way it's used right now would speed up duplicate calls during 1 page-load calculation, but would be unlikely to speed up the next page load

roddajohn:
Two comment:

I think it's currently set to 20 second cache which is not long enough for a pairing process for a tournament of, say, 80 teams.

Second, I think we can add a higher level cache that specifically caches the team / speaking ranking result page that refreshes whenever a result is changed (can be achieved by overriding the `save` method on a Round and RoundStat to set a flag of like `recalculate_stats`.  Upon thinking about it, this could help the pairings algorithm too.

And the tab_logic file is super unreadable so a higher level cache means we wouldn't have to touch it (woohoo) except maybe the cache_logic file.

roddajohn:
Will do, I'll put something real basic together.

PR #259, opened 2019-10-11
Caching results works

roddajohn:
I believe this answers #156 

Whenever the results (team or speaker) are generated, they are cached.

If a round result is changed (ie who won changed, or a speaker score / position changed) OR the force refresh button is pressed (individual to each ranking), the cache is cleared.

The makes loading the results after it has been cached pseudo-instantaneous.

### DIFF
--- a/mittab/apps/tab/debater_views.py
+++ b/mittab/apps/tab/debater_views.py
@@ -4,7 +4,7 @@ from mittab.apps.tab.forms import DebaterForm
 from mittab.apps.tab.helpers import redirect_and_flash_error, \
         redirect_and_flash_success
 from mittab.apps.tab.models import *
-from mittab.libs import tab_logic
+from mittab.libs import tab_logic, cache_logic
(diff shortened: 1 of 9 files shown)


BenMusch:
LGTM, I just need to look into memcached to make sure it's cool with caching this amount of data for this period of time. We may have to look into cutting down how much data we actually store in there (e.g. just save a sorted list of IDs rather than the entire records)

roddajohn:
Absolutely, let me know if I have to change it.

BenMusch:
Looking into it and... it seems like memcached isn't even being used lol. The cache is just stored in local memory, which is totally fine. This PR LGTM

roddajohn:
> Looking into it and... it seems like memcached isn't even being used lol. The cache is just stored in local memory, which is totally fine. This PR LGTM

That's pretty funny — usually there isn't that large of a performance difference with this size of cache, and we're on one server so the memcached multi server benefit doesn't matter.

PR #261, opened 2019-10-12
Fixed migrations

roddajohn:
Fixing the uniqueness migrations.


BenMusch:
Can you fix the build failure? It's because of the same uniqueness constraint you needed to delete from `testing_db.json`, just in `testing_finished_db.json` as well

roddajohn:
Should be good now

PR #262, opened 2019-10-13
Mass room checkins

roddajohn:
Creates the same method to use rooms as judges — ie you can check in a room per round.  None of this rank 0 nonsense — save the rank of a room and just check it out or in (all ahead of time too, woohoo!)


BenMusch:
I will get the fixtures in order later. Accidentally screwed up my local set-up so it'll take a few minutes for me to sort things out. But I believe the reason the build was failing was because the fixtures (used by the tests) weren't updated to have room check-ins, so the tests were unable to pair using the new logic

roddajohn:
Oh interesting.  Let me know if I can help.

PR #263, opened 2019-10-13
Outrounds

roddajohn:
Just sticking this up here for record keeping sake.  Will indicate when it is ready to merge (and will likely squash some commits, etc)


roddajohn:
Ok so the code here 100% works, I am now working on the publicly displayable stuff in additions to a results view in the form that tournaments are obligated to post to the forums:

[V] Ro8
XYZ (Gov) dropped to ABC

etc

I will then write up a bunch of documentation on how this works.

This DOES NOT do judge assignment, though you can use the dropdowns to hand place judges very easily as it figures out scratches for you!

roddajohn:
And there are migration problems from that unique_together issue, but I'll fix those at the end.

assets/js/outround.js
```diff
@@ -0,0 +1,168 @@
+import $ from "jquery";
```
BenMusch: IMO we should try to share the code with the normal pairings file as much as possible. I can help with refactoring JS but I definitely want to reduce duplication here before this is finalized

docs/Outrounds.md
```diff
@@ -0,0 +1,27 @@
...
+Outrounds are now here!
+
+Handling the Break
+------------------
+
+When you reach the final in-round of your tournamet, the button that is normally "Prepare Next Round" should read "Break 'em".  If this does not happen, please consult your `tot_rounds` TabSetting and ensure it is set properly.
+
+This will then break the appropriate number of teams as determined by the `nov_teams_to_break` and `var_teams_to_break` TabSetting.
+
+It will also perform a number of checks to ensure you have enough rooms and judges.  It does support paneled rounds (as long as they are consistently paneled) -- please consult the `nov_panel_size` and `var_panel_setting` for more information.
```
BenMusch: Not sure that we need to validate the panel size:

1. I think we can trust most tournaments to do this themselves

2. Sometimes panels end up being inconsistently-sized due to judging constraints and I don't want to make tournaments have to struggle with the software in that instance (For example: Dartmouth my Freshman year had a 1-person panel in novice finals)

mittab/apps/tab/judge_views.py
```diff
@@ -212,7 +212,7 @@ def batch_checkin(request):
     round_numbers = list([i + 1 for i in range(TabSettings.get("tot_rounds"))])
     for judge in Judge.objects.all():
         checkins = []
-        for round_number in round_numbers:
+        for round_number in [0] + round_numbers:
```
BenMusch: I assume `0` is how you're flagging outs? Keep in mind that when the tournament starts, `cur_round` is = `0`. Just in case that causes any issues somehow

mittab/apps/tab/migrations/0011_auto_20191013_1255.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 2.1.5 on 2019-10-13 12:55
```
BenMusch: This should be removed from this PR, right?

mittab/apps/tab/models.py
```diff
@@ -296,6 +315,71 @@ class Meta:
...
+            raise ValidationError("Chair must be a judge in the round")
+
+    def __str__(self):
+        return "Outround {} between {} and {}".format(self.num_teams,
+                                                      self.gov_team,
+                                                      self.opp_team)
+
+    @property
+    def winner(self):
+        if self.victor in [1, 3]:
```
BenMusch: nit: use the constants here to make it more readable

```suggestion
        if self.victor in [self.GOV, self.GOV_VIA_FORFEIT]:
```

mittab/libs/outround_tab_logic/bracket_generation.py
```diff
@@ -0,0 +1,16 @@
+import math
+
+def branch(seed, level, limit):
```
BenMusch: Maybe add a clarifying comment here on the variable meanings -- it's not totally clear to me on first-read

mittab/libs/outround_tab_logic/checks.py
```diff
@@ -0,0 +1,114 @@
+from mittab.apps.tab.models import *
+from mittab.libs.errors import PrevRoundNotEnteredError
+
+
+def lost_teams():
+    losers = [outround.loser.id
```
BenMusch: nit: why bother assigning this to a variable, can't we just return it inline?

mittab/libs/outround_tab_logic/checks.py
```diff
@@ -0,0 +1,114 @@
...
+
+def have_enough_judges_type(type_of_round):
+    lost_ids = [outround.loser.id
+                for outround in Outround.objects.all()
+                if outround.loser]
+
+    panel_size = 3
+
+    if type_of_round == BreakingTeam.VARSITY:
+        panel_size = TabSettings.get("var_panel_size", 3)
```
BenMusch: Because this changes as the tournament goes on, I think it may be easiest to just not have this check and let tournaments validate for themselves. Maybe still display the info in the UI but don't use it for validation

mittab/libs/outround_tab_logic/pairing.py
```diff
@@ -0,0 +1,109 @@
...
+from mittab.libs.outround_tab_logic.bracket_generation import gen_bracket
+
+from mittab.libs.tab_logic import have_properly_entered_data
+from mittab.libs import errors
+
+
+def good_to_go(num_teams):
+    # Check if people can math properly -- ie is it a power of two
+    if not math.log(num_teams, 2) % 1 == 0:
+        # ERROR: Not a power of 2, someone can't math
```
BenMusch: What about partial outrounds?

mittab/libs/outround_tab_logic/pairing.py
```diff
@@ -0,0 +1,109 @@
...
+
+    for room in rooms:
+        # THIS IS TODO
+        if not base_queryset.filter(room=room).exists() and \
+           not other_queryset.filter(room=room).exists():
+            return room
+    return None
+
+
+def whos_gov(team_one, team_two):
```
BenMusch: nit: can name this `gov_team` for consistency with the main tab logic library

docs/Outrounds.md
roddajohn: I think validating it for the like about to pair screen is still helpful.  In status quo it still lets you pair even if you don’t have enough judges so not sure what change you are suggesting — perhaps it should be more explicit that for judges it is a warning that they can safely ignore, though I still think it could help people do the math quickly.

mittab/apps/tab/judge_views.py
roddajohn: I will test to see if there are issues, I don’t think there would be though, since round 0 is never paired, it’s just the default starting value

mittab/libs/outround_tab_logic/checks.py
roddajohn: See above — the tab setting is still important as it determines how many slots are in the pairing view, I am assuming this is changed from out round to out round (and that totally works from my tests)

mittab/libs/outround_tab_logic/pairing.py
roddajohn: Currently doesn’t support, but I’ll get on that.

docs/Outrounds.md
roddajohn: In new commit, I took this validation out of the `good_to_go` method, but it remains in the UI.

mittab/apps/tab/migrations/0011_auto_20191013_1255.py
roddajohn: Done

mittab/libs/outround_tab_logic/checks.py
roddajohn: Done

mittab/libs/outround_tab_logic/pairing.py
roddajohn: Done

roddajohn:
Remaining thing to do I think

1) Combine the JS in some reasonable way
2) Partial outround support (in progress)

roddajohn:
Now that #262 is merged, I will update this to use that room check in logic.

roddajohn:
Partial outrounds are now supported, mass room check in is now supported.

mittab/apps/tab/outround_pairing_views.py
```diff
@@ -0,0 +1,639 @@
...
+            (msg, "No", "Not all rounds are entered. %s" % str(e)))
+
+    return render(request, "pairing/pair_round.html", locals())
+
+
+@permission_required("tab.tab_settings.can_change", login_url="/403/")
+def break_teams(request):
+    if request.method == "POST":
+        # Perform the break
+        backup.backup_round("before_the_break")
```
BenMusch: Let's add a human-readable timestamp in case they have to re-do the break

mittab/apps/tab/outround_pairing_views.py
```diff
@@ -0,0 +1,639 @@
...
+
+        teams, nov_teams = cache_logic.cache_fxn_key(
+            get_team_rankings,
+            "team_rankings",
+            request
+        )
+
+        nov_teams_to_break = TabSettings.get("nov_teams_to_break", 4)
```
BenMusch: What do you think about just failing the request if these aren't specified? Since it changes so much from tournament-to-tournament I feel weird about having a default here

mittab/apps/tab/outround_pairing_views.py
```diff
@@ -0,0 +1,639 @@
...
+            num_teams += 1
+
+        return redirect("outround_pairing_view",
+                        type_of_round=BreakingTeam.VARSITY,
+                        num_teams=num_teams)
+
+    pairing_released = False
+
+    if type_of_round == BreakingTeam.VARSITY:
+        pairing_released = TabSettings.get("var_teams_visible", 256) <= num_teams
```
BenMusch: I'm a bit confused by this tab setting. What's wrong with just a `pairing_released` tab setting?

mittab/apps/tab/outround_pairing_views.py
```diff
@@ -0,0 +1,639 @@
+import random
```
BenMusch: General comments on this file: Lots of pretty long view functions. Can you extract some helpers? Even if it doesn't reduce duplication, just for readability

Ex: a `get_outround_options()` function rather than appending to that variable within the view function

mittab/apps/tab/outround_pairing_views.py
```diff
@@ -0,0 +1,639 @@
...
+                       if type_of_round == BreakingTeam.VARSITY \
+                          else TabSettings.get("nov_panel_size", 3)
+    judge_slots = [i for i in range(1, judges_per_panel + 1)]
+
+    var_to_nov = TabSettings.get("var_to_nov", 2)
+    other_round_num = num_teams / var_to_nov
+    if type_of_round == BreakingTeam.NOVICE:
+        other_round_num = num_teams * var_to_nov
+
+    other_round_type = not type_of_round
```
BenMusch: IMO it's worth being a bit more verbose here for clarity:

```suggestion
    other_round_type = BreakingTeam.NOVICE if type_of_round == BreakingTeam.VARSITY else BreakingTeam.VARSITY
    ```
```

mittab/apps/tab/outround_pairing_views.py
```diff
@@ -0,0 +1,639 @@
...
+    other_round_num = round_obj.num_teams / var_to_nov
+    if round_obj.type_of_round == BreakingTeam.NOVICE:
+        other_round_num = round_obj.num_teams * var_to_nov
+
+    other_round_type = BreakingTeam.NOVICE \
+                       if round_obj.type_of_round == BreakingTeam.VARSITY \
+                          else BreakingTeam.VARSITY
+
+    excluded_judges = Judge.objects.exclude(
+        judges_outrounds__num_teams=round_obj.num_teams,
```
BenMusch: I think it might be better to not have any logic sticking judges into a particular break, since sometimes tournaments will have novices judge novice semis, but have varsity judge finals, etc. I think the benefit we get here might not be worth the trade-off

mittab/apps/tab/outround_pairing_views.py
```diff
@@ -0,0 +1,639 @@
...
+    included_judges = [(j.name, j.id, float(j.rank))
+                       for j in assign_judges.can_judge_teams(
+                           included_judges, round_gov, round_opp)]
+    included_judges = sorted(included_judges, key=lambda x: -x[2])
+    excluded_judges = sorted(excluded_judges, key=lambda x: -x[2])
+
+    return render(request, "pairing/judge_dropdown.html", locals())
+
+
+def alternative_teams(request, round_id, current_team_id, position):
```
BenMusch: We should consider hiding this in the UI, has the possibility to really screw with things if people mis-click and mess up the seeding

mittab/apps/tab/pairing_views.py
```diff
@@ -211,6 +211,8 @@ def view_round(request, round_number):
     errors, excluded_teams = [], []
     round_pairing = list(Round.objects.filter(round_number=round_number))
 
+    tot_rounds = TabSettings.get("tot_rounds", 5)
```
BenMusch: Whats the purpose of this change?

mittab/libs/outround_tab_logic/checks.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def lost_teams():
+    return [outround.loser.id
+            for outround in Outround.objects.all()
+            if outround.loser]
+
+
+def have_enough_judges_type(type_of_round):
+    lost_ids = [outround.loser.id
```
BenMusch:
```suggestion
    loser_ids = [outround.loser.id
```

Nit: imo this is a clearer name, when I read it at first I was thinking it was like the ID was missing from the db or something

BenMusch: (make sure to get the other uses of this if you commit the change)

mittab/libs/outround_tab_logic/checks.py
```diff
@@ -0,0 +1,126 @@
...
+
+    var_to_nov = TabSettings.get("var_to_nov", 2)
+
+    num_teams = teams_count
+
+    other_round_num = num_teams / var_to_nov
+    if type_of_round == BreakingTeam.NOVICE:
+        other_round_num = num_teams * var_to_nov
+
+    other_round_type = not type_of_round
```
BenMusch: Same comment as the view file w/r/t the clarity of this line

mittab/libs/outround_tab_logic/checks.py
```diff
@@ -0,0 +1,126 @@
...
+
+    num_in_use = Room.objects.filter(
+        rooms_outrounds__num_teams=other_round_num,
+        rooms_outrounds__type_of_round=other_round_type
+    ).count()
+
+    num_rooms = RoomCheckIn.objects.filter(round_number=0).count() - num_in_use
+
+    if num_rooms < rooms_needed:
+        return False, (num_rooms, rooms_needed)
```
BenMusch: To make the function usage clearer, how about we just have this return the number of rooms needed? Then we can use that logic to create the correct message in the UI. IMO function with multiple return values can make code kinda hard to read if it's not entirely clear from the name

mittab/libs/outround_tab_logic/pairing.py
```diff
@@ -0,0 +1,112 @@
...
+from mittab.apps.tab.models import *
+
+from mittab.libs.outround_tab_logic.checks import have_enough_rooms
+from mittab.libs.outround_tab_logic.bracket_generation import gen_bracket
+
+from mittab.libs.tab_logic import have_properly_entered_data
+from mittab.libs import errors
+
+
+def good_to_go(num_teams):
```
BenMusch:
```suggestion
def is_pairing_possible(num_teams):
```

Or something like that is a better function name imo

mittab/libs/outround_tab_logic/pairing.py
```diff
@@ -0,0 +1,112 @@
...
+    rooms.sort(key=lambda r: r.rank, reverse=True)
+
+    for room in rooms:
+        if not base_queryset.filter(room=room).exists() and \
+           not other_queryset.filter(room=room).exists():
+            return room
+    return None
+
+
+def gov_team(team_one, team_two):
```
BenMusch: This needs to be fixed, right?

docs/Outrounds.md
```diff
@@ -0,0 +1,27 @@
...
+========
+
+Outrounds are now here!
+
+Handling the Break
+------------------
+
+When you reach the final in-round of your tournamet, the button that is normally "Prepare Next Round" should read "Break 'em".  If this does not happen, please consult your `tot_rounds` TabSetting and ensure it is set properly.
+
+This will then break the appropriate number of teams as determined by the `nov_teams_to_break` and `var_teams_to_break` TabSetting.
```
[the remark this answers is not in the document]
BenMusch: nvm, you did that

docs/Outrounds.md
```diff
@@ -0,0 +1,27 @@
...
+
+It will also perform a number of checks to ensure you have enough rooms and judges.  It does support paneled rounds (as long as they are consistently paneled) -- please consult the `nov_panel_size` and `var_panel_setting` for more information.  It will let you pair if you don't have enough judges, but it will warn you.
+
+The other tab setting that must be set correctly in order to ensure judges / rooms are not double booked is the `var_to_nov` variable.  If you would like varsity octafinals to happen at the same time as novice quarterfinals, this value should be `2` as the quotient of number of teams in varsity break rounds to the simultaneous novice break round is 2.  If they are happening at the same time, the value should be `1`, if they are octafinals at the same time as semifinals, then it should be 4, etc.
```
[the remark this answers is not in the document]
roddajohn: Currently this is not supported.

I'll look into adding that flag.

mittab/apps/tab/forms.py
```diff
@@ -90,9 +90,12 @@ def __init__(self, *args, **kwargs):
                 checkins = [
                     c.round_number for c in CheckIn.objects.filter(judge=judge)
                 ]
-                for i in range(num_rounds):
+                for i in range(-1, num_rounds):
```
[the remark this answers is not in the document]
roddajohn: Comment included — out round number as a global constant would be awkward due to how many of the functions use i + 1, so the loop would have to start at `OUT_ROUND_NUMBER` - 1.  I'll just try and comment everywhere it appears.

mittab/apps/tab/forms.py
```diff
@@ -593,3 +596,66 @@ def score_panel(result, discard_minority):
...
+        round_obj.victor = cleaned_data["winner"]
+        round_obj.save()
+
+        round_obj = Outround.objects.get(pk=cleaned_data["round_instance"])
+        if round_obj.victor > 0:
+            winning_team_seed = round_obj.winner.breaking_team.effective_seed
+            losing_team_seed = round_obj.loser.breaking_team.effective_seed
+
+            if losing_team_seed < winning_team_seed:
+                round_obj.winner.breaking_team.effective_seed = losing_team_seed
```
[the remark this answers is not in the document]
roddajohn: This isn't used for final rankings, it's used for determining pairings.  The "Forum View" would show final results and gives a "xyz dropped to abs in Octos" etc.

The seeding functions because if the 8 beats the 1, they are paired where the 1 would've been paired in a quarters break.  They basically get the 1st seed to progress through outs.

mittab/apps/tab/pairing_views.py
roddajohn: This is so that on the pairings view the button can change to do the break when we are at the final round.  The template didn't know what the total number of rounds was before.

mittab/libs/outround_tab_logic/checks.py
roddajohn: I was just following the logic from the other `tab_logic.py` I can totally change to be better though.

mittab/libs/outround_tab_logic/pairing.py
roddajohn: Supports partials now!

mittab/libs/outround_tab_logic/checks.py
roddajohn: Done

mittab/libs/outround_tab_logic/pairing.py
roddajohn: So here's my thoughts on this:

There's a bunch of logic here with regard to the method that different tournaments do outrounds (flip for choice, least choice chooses, flip for side, sidelocked or not, etc).  I think the best way to implement this is through at least a `SIDELOCK` `TabSetting`.  There should then be a toggle for each outround that is like are these sides real, or are they placeholders to be changed after coin flip, etc.  Then the display will change.  Been thinking about the best way to do that (I don't plan on having it keep track of choice, etc).

Thoughts ?

mittab/apps/tab/outround_pairing_views.py
roddajohn: With these two tab settings there are a few benefits:

1) I can track whether varsity / novice are released separately (some tournaments will split them up if one is running late, etc)
2) The pairings view isn't for one round, but rather for all of outs (this is a nice 1 screen record for outrounds), thus if you just have a toggle, it's unclear which round you are showing and hiding

mittab/apps/tab/outround_pairing_views.py
roddajohn: I think I got all instances of this in recent commit

mittab/apps/tab/outround_pairing_views.py
roddajohn: This logic doesn't stick any judges in a certain break, it just ensures that judges who are excluded are really excluded (ie not judges a novice round happening at the same time, etc).  This is just meant to allow a TD to use the interface a little bit faster, all checked in outround judges should show up, however, in the included_judges variable if they're paired in to either the varsity, or novice round.

mittab/apps/tab/outround_pairing_views.py
roddajohn: Important for switching gov and opp though.  I think this issue should exists (someone could manually mess up power pairing for example).  Better to give them the option and have them revert from backup if they really fuck up.  However, since it shows seeding in team cards, it's not that hard to fix.

mittab/apps/tab/outround_pairing_views.py
roddajohn: Done

mittab/apps/tab/outround_pairing_views.py
roddajohn: Done

mittab/apps/tab/outround_pairing_views.py
roddajohn: Added a `get_outround_options` function.  By way of how the view functions tend to be setup in `pairing_views.py` i'm not sure whether I should write it basically from scratch better, or leave it how it is now to be consistent.

docs/Outrounds.md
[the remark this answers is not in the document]
roddajohn: Done

roddajohn:
Current outstanding things:

1) Refactoring JS (lol I hate JS)
2) Maybe reworking some of the `checks.py` code (I did a little, and reworked some views code)
3) Figure out government assignment algorithms / options (working on this now)
4) Unittests (this is very close to done)

mittab/libs/outround_tab_logic/pairing.py
BenMusch: I see. This seems like a reasonable placeholder.

BenMusch:
@roddajohn I can work on the JS, don't worry about that

roddajohn:
Ok I think this is at a good spot.

@BenMusch any thoughts as to why the `import_team` tests are failing ?

Outround tests are working it looks like.

Any other work to be done on this PR?

roddajohn:
Public record of test:

I spoofed an entire tournament (both with flip for choice, sidelock, no sidelock, flip for side, etc) and it all works!  Woohoo!

BenMusch:
@roddajohn The team stuff is failing because this change breaks the import team code now that the `break_preference` field is required, I think. Maybe the default isn't propagating to the `TeamForm` correctly? I'll try to debug

BenMusch:
Yeah looks like that was it, this commit should fix it

BenMusch:
Played around with this a bit locally, biggest thing that comes to mind for me would be adding the option to "offset" novice outs by a round or two. For example, tournaments running octofinals will often hold novice semis during varsity quarters, otherwise the judge requirements are too high

roddajohn:
Maybe I'm missing something, but I think in this case you should be able to set `var_to_nov` to be equal to 4 (16 teams in VRo16 and 4 in NRo4, thus the quotient is 4).  It will generate the novice break when you click break 'em but, you need not release these pairings till you run the VRo16.

Does this make sense?  Perhaps this should be explained better in the docs.

BenMusch:
@roddajohn I see, I think I misunderstood it. Maybe we could switch to an absolute offset? That seems to be the more straightforward way to think about it

roddajohn:
Changed it to an offset from a quotient.

mittab/libs/tests/tab_logic/test_outround_pairing.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def generate_checkins(self):
+        for r in Room.objects.all():
+            RoomCheckIn.objects.create(room=r,
+                                       round_number=0)
+
+    def test_break(self):
+        self.generate_checkins()
+
+        outround_tab_logic.perform_the_break()
```
BenMusch: Ideally we would be testing more than just that the code runs without errors -- we should be validating the pairings generated. Also would be nice to have:

1. A test which iterates through the outrounds, inputting results and confirming the correct results of the next round

2. One which starts with partials

roddajohn: Yep, I’ll get on that — I’m always lazy with unit tests haha

roddajohn: Newest commit does both these things.

roddajohn:
Todo (based on Penn, in progress):

All non scratched judges should be available (even if they judges before in the tournament)

roddajohn:
Also generate scratches (in case there is a new judge)

BenMusch:
@roddajohn +1 -- scratch generation should be done on-create for judges, in addition to before each round.

roddajohn:
Fixed the non scratched judges showing up in the dropdowns (ignoring previous pairings).

Generate scratches I'll generate a new issue for, but I will add this upon pairing each outround.

BenMusch:
BTW, does this let judges judge teams again in outs?

roddajohn:
Yes, this allows repeat judging in outs.

roddajohn:
Used at AU ProAms with success, few comments:

1) If a judge is added for in rounds, scratches may not be generated properly
2) Change display of the outround pairings to be in order of seed to aid in judge selection
3) Reinvestigate various outround choice / least gov mechanisms

All are from Sergio, Caleb said he had thoughts, which I haven't yet gotten

mittab/apps/tab/api.py
```diff
@@ -0,0 +1,146 @@
+import json
+
+from mittab.apps.tab.models import *
+from mittab.libs import tab_logic
+
+
+def get_num_rounds_debated(team):
```
BenMusch: Is this for standings site integration? If so, can you move it to another PR?

roddajohn: #whathappenswhenImessupwithgit

Yeah absolutely

PR #272, opened 2019-10-21
Fixes 271

### LINKED ISSUE #271: Judge school-based scratches stick if they chance schools
Example:

1. Someone just is accidentally assigned to the wrong school in R1 (imagine someone confuses Wellesley and Wesleyan, for example)

2. R1 is paired with the wrong judge affiliation

3. The affiliation is removed before R2

4. R2 is paired

Expected behavior: The judge should have no scratches based on the old school assignment

Actual: All of the scratches stay the same

Fix:

1. We need to distinguish between normal scratches and school affiliation ones, so we can remove them, so we should add another scratch type

2. Every time a judge is saved, re-calculate the school scratches

roddajohn:
Resolves #271


BenMusch:
LGTM, with one other case I realized: This should also happen when the team are saved if the teams for that school switch

Issue #276, opened 2019-11-02
Fix whatever tf just happened


BenMusch:
Sqlite3 was corrupted?? Idk what happened there

But, lots of things should be improved to prevent worst-case data lock:

1. Holy shit holy shit holy shit let's not use sqlite3 anymore

2. Store backups more frequently (save _after_ judge assignment also)

3. Switched to managed postgres?

PR #277, opened 2019-11-02
sqlite->MySQL, implement better backup mechanism

BenMusch:
Fixes https://github.com/MIT-Tab/mit-tab/issues/276


BenMusch:
Brief update: Gonna hold off on using RDS and instead run mysql in docker for now. Since the issue was sqlite, it seems like overkill to correct for the local backups which weren't part of the problem. But I'm gonna keep this refactoring so that if we need to backup in more ways in the future it's possible

ifly6:
I have two question about this.

1. Why do you think SQLite is bad?
2. This seems to be a response to something, what happened?

BenMusch:
sqlite isn't bad, its use-case is just very un-scalable.

At both NU and AU proams this weekend, there was an issue where an error was thrown with `database disk image is malformed` which basically takes down the entire server and can potentially cause data loss

Basically every single post I found about how to trouble-shoot the error was about manual operations you can do to recover the data. But nothing about a root cause and how to prevent it from happening. Since I can't find anything available on how to fix the problem, and manually recovering data from every tournament where this happens is not a reasonable solution, it seems like the only real option here is to stop using sqlite

BenMusch:
There are also other benefits which come from shifting away from sqlite. Most notably, that's the biggest barrier that prevented mittab from being hosted on heroku. If we can get it hosted on heroku, things become basically free and my life gets much easier

roddajohn:
Another issue with sqlite in general is that there's a db concurrency issue which generates /a lot/ of crash reports.  This is totally a good change.

PR #299, opened 2020-09-10
Discord Integration

roddajohn:
This is totally in progress.

Things todo as of now:
- [x] Add a Discord documentation page
- [x] Add more useful messages in Discord with regard to the commands available


roddajohn:
I think this is ready:

Working on the deployer now.

roddajohn:
Actually, let's hold off for at least a few hours as I figure out the docker-compose stuff related to the deploy.

Dockerfile.discord~
```diff
@@ -0,0 +1,22 @@
+FROM python:3.7
```
BenMusch, line 1: What's the difference between this file and `Dockerfile.discord`? Do we need both?

mit_tab_bot/api.py
```diff
@@ -0,0 +1,64 @@
...
+    return api
+
+
+def get_judge(discord_id):
+    api = construct_api()
+
+    response = api.judges.list(
+        params={'discord_id': discord_id}
+    )
+    return response.body[0] if len(response.body) > 0 else None
```
BenMusch, line 27: nit: can just be `if response.body` or `if len(response.body)`

mit_tab_bot/roles.py
```diff
@@ -0,0 +1,117 @@
...
+
+from .api import (
+    is_judge,
+    is_debater,
+    get_judge,
+    get_debater
+)
+
+ADMIN = [
+    'Talos#5956'
```
BenMusch, line 11: Can you explain what exactly this corresponds to? Is this your personal user id or something?

BenMusch (commented):
High-level comment: What do you think about shipping `mit_tab_bot/*` as a separate repo? It can build and public to dockerhub and then we would reference that in `docker-compose`

roddajohn:
@BenMusch on the high level: yup that makes sense I'll start working on this, everything else is inline

mit_tab_bot/api.py
roddajohn: lol yes

mit_tab_bot/roles.py
roddajohn: Yeah this is my personal Discord ID which grants superuser, plan is to make this part of the form on the deployer, and passed through an ENV variable, but I haven't finished the deployer piece yet so here it is.

BenMusch: Gotcha, yeah would be nice to have this set via some env var/config file and then referenced by a constant like `ADMIN_DISCORD_ID` to make things clear

PR #309, opened 2021-04-08
Use a profiler, reduce DB queries

BenMusch:
We are literally doing thousands of queries for some page loads

This PR:

 - Adds the `silk` package for profiling, which can be turned on with the `USE_SILK` env var (set to any non-empty value)
 - Refactors a lot of tab calculation code to pre-load all relevant data and then do calculations in-memory (rather than on the DB)
 - ~~Refactors TabSettings to cache all settings, and invalidate them when a write happens~~
   - **TODO** Need to test this with multiple gunicorn workers, not clear if this works in prod
     - ~~This would have broken in prod. May be worth refactoring later. Until then, the tabsettings queries are not enough of an issue to block this PR. If we want cross-process caching, we can use redis/memcached in the future~~
     - Okay, this was actually pretty necessary. We can use the local file system cache that django provides to provide cross-process consistency, and then move to redis in the future


mittab/apps/tab/models.py
```diff
@@ -3,21 +3,16 @@
 from haikunator import Haikunator
 from django.db import models
 from django.core.exceptions import ValidationError
-from polymorphic.models import PolymorphicModel
 
+from mittab.libs import cache_logic
 
-class ModelWithTiebreaker(PolymorphicModel):
```
BenMusch, line 9: For some reason django-polymorphic introduces some bad behavior with prefetch_related, since it's such a minor thing here I chose to just remove the dependency

BenMusch:
Seems like most of these changes have been implemented, not sure why this is still open

PR #320, opened 2021-10-02
Tab-Policy: fix typo regarding byes

DanielS6:
Following discussion of the first round, the next bullet point is likely intended to explain the policy all subsequent rounds rather than all previous rounds, so use "succeeding" rather than "preceding".


BenMusch:
thanks @DanielS6 !

PR #338, opened 2023-02-26
Fixing an issue with judge assignment introduced in Feb 2022

roddajohn:
This is preventing judge assignment from working for non R4s...


roddajohn:
Relevant commit: https://github.com/MIT-Tab/mit-tab/commit/14c6f4907bc208b1d17f0fca83e8953643a6e736

BenMusch (approved):

PR #339, opened 2023-02-26
Fixing Rehits

roddajohn:
Been seeing a lot of rehits in recent tournaments, this fixes that bug


mittab/libs/tab_logic/__init__.py
```diff
@@ -350,7 +351,7 @@ def hit_before(team1, team2):
         if round_obj.opp_team == team2:
             return True
     for round_obj in team1.opp_team.all():
-        if round_obj.opp_team == team2:
+        if round_obj.gov_team == team2:
```
roddajohn, line 354: This is the substantive change

BenMusch (approved):

PR #347, opened 2024-12-21
Room Dropdowns (Room Tags Part 1/n)

JoeyRubas:
First PRs leading up to the room tag feature set.

Includes the following loosely associated features/changes:

**Big features and changes:**
1.  **Room dropdowns:** Add room dropdowns to pairing - should be much more convenient than only being able to change rooms in admin
2. **Dedicated "pair rooms" button:** Migrate room pairings out of the main tab logic into a dedicated separate function with MWM:
- the migration decision was motivated by tab staff likely needing to re-run room pairing (i.e, forgot to mark any rooms as accessible; they shouldn't then be forced to do manual room pairing) and that the more complicated pairing logic introduces more room for error that I thought needed to be separated from just pairing logic (both referring to server-side errors and "not enough accessible rooms" errors
- MWM is, of course, massively overkill for the non-room-tag logic, but even with three tags, there are a lot of situations where the naive/greedy room assignment algorithms fail, so this seems worthwhile.

**Small Changes:**
1. Add a setting to disable room pairing by speaks - APDA seems to want the option

Edit: Removed the bloated extra features that are no longer in the PR from the PR description


mittab/apps/tab/pairing_views.py
```diff
@@ -105,6 +107,100 @@ def assign_judges_to_pairing(request):
...
+            pass
+
+    # Fetch all rooms checked in for the given round, ordered by rank
+    # A little out of depth on the Django ORM here, but this seems like the
+    # way to get this data with the least amount of queries
+    rooms = set(Room.objects.filter(
+        roomcheckin__round_number=round_number
+    ).annotate(
+        has_round=Exists(Round.objects.filter(room_id=OuterRef('id')))
+    ).order_by("-rank"))
```
BenMusch, line 146: The `set()` is for the subtraction operator, right?

I'd prefer something more like:

```python3
viable_unpaired_rooms = []
viable_paired_rooms = []
other_rooms = [] # populate this if needed, otherwise delete

for room in rooms:
    if room.has_round:
        viable_paired_rooms.append(room)
    else:
        viable_unpaired_rooms.append(room)
```

mittab/apps/tab/pairing_views.py
```diff
@@ -105,6 +107,100 @@ def assign_judges_to_pairing(request):
...
+            "success": True,
+            "room_color": room_color if room_color else "",
+            "room_id": room_obj.id,
+            "round_id": round_obj.id,
+            "room_name": room_obj.name,
+        }
+    except Exception:
+        emit_current_exception()
+        data = {"success": False}
+    return JsonResponse(data)
```
BenMusch, line 202: nit: if we can make this response have an error code (as opposed to returning a 2XX) it'll be slightly more idiomatic

mittab/apps/tab/pairing_views.py
```diff
@@ -105,6 +107,100 @@ def assign_judges_to_pairing(request):
...
+    return render(request, "pairing/room_dropdown.html", {
+        "current_room": current_room_obj,
+        "round_obj": round_obj,
+        "viable_rooms": viable_rooms,
+        "viable_paired_rooms": viable_paired_rooms,
+        "other_rooms": other_rooms,
+    })
+    
+@permission_required("tab.tab_settings.can_change", login_url="/403/")
+def assign_room(request, round_id, room_id, remove_id=None):
```
BenMusch, line 183:
```suggestion
def assign_room(request, round_id, new_room_id, prev_room_id=None):
```

This naming is a bit more intuitive to me, assuming I'm parsing how this is used correctly

mittab/apps/tab/models.py
```diff
@@ -382,6 +382,23 @@ def delete(self, using=None, keep_parents=False):
...
+    
+    def get_color(self):
+        """Colors are inherited from the highest priority room tag
+        
+        Not sure if the typical approach would be method or property
+        but since room color can change both from changes to any room tag,
+        or changes to which tags are associated with this room
+        this seemed like it would probably be the lighter weight approach,
+        especically because I expect num_room_tags <<<<<< num_rooms in virtually every case    
+        """
```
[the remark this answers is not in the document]
BenMusch: Either way, this is the sort of thing that I bet we can calculate best by creating some set of valid colors and then deterministically mapping something static (e.g. priority level, room name, room ID, etc.) to that list of colors

mittab/libs/assign_rooms.py
```diff
@@ -0,0 +1,65 @@
...
+            weight -= room.rank
+
+            edge = (pairing_i, len(pairings) + room_i, weight)
+            graph_edges.append(edge)
+    
+    #This is of course super overkill for the required logic in this branch
+    # but even with just ~3 room tags it was pretty easy to find scenarios 
+    # where the simpler/more naive greedy algorithms would fail. The overhead
+    # isn't too bad, (it was under 0.02 seconds on my system) 
+    # and since we already use this elsewhere it seemed reasonable
```
BenMusch, line 47: I would use this code comment to document an example of an instance where a simpler alg fails rather than benchmarks

mittab/libs/assign_rooms.py
```diff
@@ -0,0 +1,65 @@
...
+
+            #High seed high room bonus
+            if no_seeding == 0:
+                weight -= abs(pairing_i - room_i)
+
+            #Bad room penalty
+            weight -= room.rank
+
+            edge = (pairing_i, len(pairings) + room_i, weight)
+            graph_edges.append(edge)
```
BenMusch, line 41: This sort of match alg isn't intuitive to me, but I can see how it could work. Would the idea be to score the matching based on the specific pairing it's being assigned to?

e.g. With these two pairings:
 - Top-seeded round with debater(s)/judge with a wheel chair
 - Bottom-seeded team with no accessibility requirements

And these rooms:
 - Room in GA building, but not wheelchair accessible
 - Room a few minutes away, but is wheelchair accessible

We'd want the weight of the matching to be dynamic, where the wheelchair accessible room is the top match for the first round but not the second

BenMusch (commented):
first pass

mittab/apps/tab/models.py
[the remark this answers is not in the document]
JoeyRubas: This was the intended functionality, where tab staff can assign different colors to each room tag, and then rooms are colored in by their highest priority tag. I hadn't thought of just limiting to a set few colors but that's probably the more reasonable/less overkill way to go, do you think there's a lightweight way to keep the color selector or is limited set of colors the way to go?

Screenshot 2024-12-26 115737

image

image

mittab/libs/assign_rooms.py
JoeyRubas: Yeah, pretty much, especially for the "high seed high room" bonus, which replicates the existing functionality of trying to give the n-ranked room to the n-ranked round.

For tags, the idea isn't necessarily that accessible rooms will get bad weights for debaters who don't need them as much as just putting a substantial negative weight on unfulfilled room tags, meaning max weight matching will first do whatever it can to fulfill all room tags (and to fulfill the higher weight ones first), and only then look at the smaller priorities like prioritizing good rooms, and giving better teams better rooms.

This is roughly what that will look like:
```
            #Unfilled tags penalty
            room_tags = room_tags_dict[room.id]
            unfulfilled_tags = pairing_tags - room_tags
            if unfulfilled_tags:
                unfulfilled_tags_matrix[pairing_i][room_i] = unfulfilled_tags

                #Arbitrary weight that's large enough to gaurentee precedence
                weight -= 1000 * sum(tag.priority for tag in unfulfilled_tags) 
```

mittab/apps/tab/models.py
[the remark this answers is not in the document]
BenMusch: I'd suggest offering a list of ~8 colors initially, highly unlikely more than that is needed

mittab/apps/tab/migrations/0022_auto_20241221_2217.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('tab', '0021_auto_20231104_1847'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='round',
+            name='room',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, to='tab.Room'),
```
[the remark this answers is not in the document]
BenMusch: ^ bumping this

mittab/libs/assign_rooms.py
```diff
@@ -0,0 +1,64 @@
+import random
+from django.db import transaction
+from mittab.apps.tab.models import RoomCheckIn, Round, TabSettings
+from mittab.libs import errors, mwmatching, tab_logic
+
+
+def add_rooms():
+    no_seeding = TabSettings.get("disable_room_seeding", 0)
```
BenMusch, line 8: nit: I'd prefer the tab setting to be `enable_room_seeding` so that the code is more intuitive, e.g. `true` == `enabled` is more intuitive than `true` == `disabled`

mittab/libs/assign_rooms.py
```diff
@@ -0,0 +1,64 @@
...
+    rooms = RoomCheckIn.objects.filter(
+        round_number=round_number).select_related("room")
+    rooms = sorted((r.room for r in rooms), key=lambda r: r.rank, reverse=True)
+    pairings = tab_logic.sorted_pairings(round_number)
+
+    if no_seeding:
+        random.shuffle(pairings)
+
+    if not pairings or not rooms or len(pairings) > len(rooms):
+        raise errors.RoomAssignmentError("Not enough rooms or pairings")
```
BenMusch, line 24:
```suggestion
    if not pairings:
        raise errors.RoomAssignmentError("Attempted to assign rooms to an empty pairing")
    if len(pairings) > len(rooms):
        raise errors.RoomAssignmentError(f"Not enough rooms. Found {len(pairings)} rounds and only {len(rooms)} rooms")
```

BenMusch: Current error handling was collapsing two different failure conditions into one, this will make it a touch easier to see what goes wrong

mittab/libs/assign_rooms.py
```diff
@@ -0,0 +1,64 @@
...
+
+    graph_edges = []
+
+    for pairing_i, pairing in enumerate(pairings):
+        for room_i, room in enumerate(rooms):
+            weight = 0
+
+            # High seed high room bonus
+            if no_seeding == 0:
```
BenMusch, line 33: let's move the == 0 check to where we instantiate no_seeding, it's more intuitive to make that a bool right away

mittab/libs/assign_rooms.py
```diff
@@ -0,0 +1,64 @@
...
+                weight -= abs(pairing_i - room_i)
+
+            # Bad room penalty
+            weight -= room.rank
```
BenMusch, line 37: I don't fully understand the value of a bad room penalty here. We're already checking the relative mis-match with the pairing priority above

Is the idea here to account for scenarios where there are more rooms than pairings, and therefore we want an additional penalty for the worst rooms, to make sure they're de-prioritized? If so:
 1. Can we get away with just ignoring the bottom-ranked rooms?
 2. If we can't (perhaps this wouldn't work with room tags), I think we may wanna make this penalty even stronger. 1 point of room rating feels more significant than 1 level of mis-match in the pairing priority, and right now they're weighted equally. I'd suggest `-= room.rank * 100`

mittab/libs/assign_rooms.py
```diff
@@ -0,0 +1,64 @@
...
+
+            edge = (pairing_i, len(pairings) + room_i, weight)
+            graph_edges.append(edge)
+
+    # This is of course super overkill for the required logic in this branch
+    # but even with just ~3 room tags it was pretty easy to find scenarios
+    # where the simpler/more naive greedy algorithms would fail. The overhead
+    # isn't too bad, (it was under 0.02 seconds on my system)
+    # and since we already use this elsewhere it seemed reasonable
```
BenMusch, line 46:
```suggestion
```

Since this comment is specific to this diff, no need to merge it to `main`

mittab/libs/assign_rooms.py
```diff
@@ -0,0 +1,64 @@
...
+    room_assignments = mwmatching.maxWeightMatching(
+        graph_edges, maxcardinality=True)
+
+    if -1 in room_assignments[:len(pairings)]:
+        pairing_list = room_assignments[: len(pairings)]
```
BenMusch, line 51:
```suggestion
        pairing_list = room_assignments[:len(pairings)]
```

mittab/libs/backup/__init__.py
```diff
@@ -44,6 +44,9 @@ def backup_round(key=None, round_number=None, btime=None):
         if round_number is None:
             round_number = TabSettings.get("cur_round", "no-round-number")
 
+        elif round_number == 0:
```
BenMusch, line 47: Is this meant to be a part of a different work related to the backup/restore issues?

mittab/libs/errors.py
```diff
@@ -55,3 +59,15 @@ def __init__(self):
...
         return repr(self.msg)
+
+
+class RoomAssignmentError(Exception):
+    def __init__(self, reason=None):
+        super(RoomAssignmentError, self).__init__()
+        if reason is not None:
+            self.msg = reason
+        else:
+            self.msg = "Could not assign judges"
```
BenMusch, line 70:
```suggestion
            self.msg = "Could not assign rooms"
```

BenMusch (commented):
Code looking good! But unable to test this out locally, any idea what's going on? https://www.loom.com/share/d844d73648644c40b17ba97becd1558f

BenMusch (commented):
Code looking good! But unable to test this out locally, any idea what's going on? https://www.loom.com/share/d844d73648644c40b17ba97becd1558f

JoeyRubas:
> Code looking good! But unable to test this out locally, any idea what's going on? https://www.loom.com/share/d844d73648644c40b17ba97becd1558f

seems like I made a very silly error in January (that I'm very surprised I didn't test), where when I removed the warnings return from .add_rooms() (will re-visit that when I get to room tags), I for some reason deleted the whole line 
`warnings = assign_rooms.add_rooms()`
which of course meant add_rooms() wasn't getting called at all. Should be back to working now, apolgies for the silly commit

JoeyRubas:
Missed some bumps from earlier in the month but should be resolved now. I also removed the unrelated features and one-liners that really shouldn't be in here, along with like 200 lines of formatting artifacts.

BenMusch:
Sounds good! I hopefully will take another pass at review this weekend

BenMusch (approved):

PR #348, opened 2025-01-16
Setup Documentation

JoeyRubas:
Apologies for so many concurrent PRs, but I was setting up mit-tab on my new computer to proceed with the previous PRs, and I decided to document my process to generate a more robust installation guide. 

After I wrote the guide, I set up a new WSL Ubuntu instance for verification, and was able to get the dev server running from scratch in under 5 minutes - although ideally, I'll have some others run through the guide on their systems to verify success across systems before merging.


BenMusch:
ty! Will probably take a pass at this myself to set-up my new laptop before merging

docs/setup.md
BenMusch, line 1: at a high-level, I think we can keep this to the README (or a separate file that doesn't get picked up by the docks

BenMusch: Also should probably be called something like `WINDOWS_DEVELOPMENT.md`

docs/setup.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Step 5: Set Up the Python Environment
+
+### Install Python 3.7.13 Using Pyenv
+
+```bash
+pyenv install 3.7.13
+pyenv local 3.7.13
+```
+
```
BenMusch, line 133: IMO this section is too specific:
 - no need to provide pyenv as the single solution, we can link to it as an example
 - We shouldn't hard-code how to install pyenv, we should link to their docs and count on them keeping it up to date (in case installation instructions change, e.g.)
 - No need to hard-code a python version, as it's tracked in `.python-version` in this repo

docs/setup.md
```diff
@@ -0,0 +1,194 @@
...
+```bash
+python -m venv venv
+source venv/bin/activate
+pip install pipenv
+pipenv install --python 3.7
+```
+> **Note:** For unknown reasons, `pipenv install --python 3.7` occasionally fails the first time it is ran, but passes if ran a second time.
+
+---
```
BenMusch, line 144: I think venv and pipenv are kinda overlapping in their purpose here. Need to look into the current way to do things but venv should not be needed

docs/setup.md
```diff
@@ -0,0 +1,194 @@
...
+
+4. Have fun!
+Above all else, the great benefit of an open-source tab site is it gives **you** control to add features and functionality you or your league desire! Let this be your opportunity to add features that make your league better, and if you're lucky, you might just learn some software skills along the way :)
+
```
BenMusch, line 194: Should be separate from the set-up guide file, but not oppossed to these inclusion

docs/setup.md
```diff
@@ -0,0 +1,194 @@
...
+1. Linting:
+The GitHub is configured to automatically run `PyLint` and check for errors. It is recommended that you install a code-formatter like `black` to take care of the more menial formatting details (especially whitespace and line length), and PyLint to check your code for linting errors before opening a PR. Integration with your IDE will vary, but if you'd like to use this just in terminal, you can do so with 
+
+```bash
+pip install black, pylint
+```
+
+2. Versions:
+While this may change in the coming months, at the time of writing many of the libraries and tools used in this repo are somewhere between 5-10 years behind their current version. For example, the app runs on Django version `2.2.10`. Keep this in mind when searching for documentation online, since parsing what advice applies to what version can occasionally be a headache
```
BenMusch, line 187: Would stay away from hard-coding any versions, we can just reference that the code is old and to make sure you're looking at the right docs version

docs/setup.md
```diff
@@ -0,0 +1,194 @@
...
+
+3. Django ORM 
+Django uses a clever API to hide database operations from the developer and allow you to interact with the database through their Object Relational Mapper (ORM), which reads more like simple Object Oriented programming than database operations. While this is a major feature and convenience in many ways, it gives way to many pitfalls for rookie developers (including myself circa ~2 months ago at time of writing), since it makes it easy to accidentally write code that makes **thousands** of database calls unintentionally. Given this, make sure to read a basic guide on the Django ORM [(like this one)](https://medium.com/@RohitPatil18/n-1-problem-in-django-and-solution-3f5307039c06), and pay special attention to preventing the `N+1` problem, by using methods like `prefetch_related`. Lastly, make sure to profile your code with[`silk`](https://medium.com/@sharif-42/profiling-django-application-using-django-silk-62cdea83fb83) to verify you're not accidently overloading the database.
```
BenMusch, line 190: Let's avoid first-person, key points to hit are:
 - Link to the profiler
 - Link to N+1 problem
 - Adding info about how to log DB queries in django (I think there's a setting, can look into this)

docs/setup.md
JoeyRubas: This makes sense, although my hope was to make the installation process as close to a copy paste as possible to keep the barrier to entry lower. I wonder if there's a good meet in the middle, how would you feel about something like:


### Install Pyenv

Install Pyenv with the following command:

```bash
curl -fsSL https://pyenv.run | bash
```
>**Note:** Pyenv is just one of many potential choices for this step

Follow the instructions outputted by this command on how to add pyenv to your shell (this should be a few terminal lines you can copy and paste). Then, restart your shell.

---

## Step 5: Set Up the Python Environment

### Install Python 3.7.13 Using Pyenv

Check `.python-version` for the most up to date python version number. Then run
```bash
pyenv install [version number]
pyenv local [version number]
```

docs/setup.md
JoeyRubas: moved back to root dir, and renamed as suggested.

docs/setup.md
JoeyRubas: fixed

docs/setup.md
JoeyRubas: Re-wrote to remove versions, first person, and get a bit closer to a documentation writing style.

ianmcvann:
Just added the MacOS documentation to the PR.

docs/setup.md
ianmcvann: It's not, fixed in both the windows docs and macos.

JoeyRubas: Re: this, gonna spend some time testing and verifying tonight but my understanding was

1. when you create a pipenv, it also creates and manages the a venv, and if it's already in a venv it will use the one that already exists, so this is functionally the same just with a different order of operations 
2. Wsl creates some issues if you don't manually create the venv first, because it wants you to use sudo apt-get for global python package installs which was being finnicky for me. 

Seems definitely unneeded for macos though

PR #351, opened 2025-01-25
Version Migrations

JoeyRubas:
Big version migration branch. Seems hard to do (at least some) these individually because so many are inter-dependent (eg. django-webpack loader depends on python, Django, node, and webpack versions, but after I get everything working, I'm happy to try to break up into singleton version migration PRs as best I can. Very rough around the edges atm, but putting up as a draft for visibility (wanted to send it to Ian as a PR).

Specific migrations are:
1. Python 3.7 -> 3.12
2. Django 2.2 -> 5.0
3. Node 18 -> 22
4. misc packages, including some swap outs of deprecates for replacements (i.e. raven -> sentry SDK)
5. some tidying meant to get rid of warnings that show up every time you run the dev server. i.e: some packages get imported many times from different places like jquery, the entire jquery library was getting packaged up by webpack which causes a constant "webpack too big" warning, etc

Note that pipenv can be a bit weird when migrating python versions, because it is written in python, and depends on your python version. Can cause some weirdness.


BenMusch (commented):
Are any of these migrations dependent on one another? would be simplest to merge one at a time if possible

JoeyRubas:
@BenMusch  Gonna meet with ian and see if we can come up with something, but TLDR: there is a good bit of messy interdependence, and I think Django and python need to be upgraded in tandem unless we add an additional intermediate python version.

> Django 2.2 (what we use now) only supports up to Python 3.10
> Django 5.0 requires at least 3.10
> Django-webpack-loader has no versions that support Python 3.10 and Django 2.2
> Django 4.2 (oldest LTS version) requires at least 3.8

Best answer is probably something like migrate to python 3.10 and django 4.2 in one go, probably future proof enough to prevent another major upgrade for at least 1.5 yrs.

PR #354, opened 2025-02-25
Add a toggle button to enable viewing member names of teams in pairings

Speedrace4:


assets/js/displayMemberNames.js
```diff
@@ -0,0 +1,35 @@
+import $ from "jquery";
+
+$(document).ready(() => {
+  $('#names').click(function() {
+    changeDisplayNames();
+  });
```
BenMusch, line 6: nit:
```suggestion
  $('#names').click(changeDisplayNames);
```

assets/js/displayMemberNames.js
```diff
@@ -0,0 +1,35 @@
...
+});
+
+function changeDisplayNames() {
+  const teamNamesCurrentlyDisplayed = !document.getElementById("names").checked
```
BenMusch, line 10: Let's use a more specific ID (obviously DOM will need to be updated too)

```suggestion
  const teamNamesCurrentlyDisplayed = !document.getElementById("name_display_toggle").checked
```

assets/js/displayMemberNames.js
```diff
@@ -0,0 +1,35 @@
...
+      for (let i = 0; i < teamNames.length; i++) {
+          const element = teamNames[i];
+          element.style.display = "none";
+      }
+      const memberNames = document.getElementsByClassName("memberNames")
+      for (let i = 0; i < memberNames.length; i++) {
+          const element = memberNames[i];
+          element.style.display = "block";
+      }
+  }
```
BenMusch, line 34: Rather than iterating through and manually applying styles, I think a better approach would be: Apply a class to the wrapper element like `team-name-card team-name-visible`. When the checkbox is toggled, replace `team-name-visible` to `team-name-hidden`, and in the CSS set `display: block`/`display: none` based on the class

mittab/apps/tab/pairing_views.py
```diff
@@ -414,6 +414,7 @@ def pretty_pair(request, printable=False):
 
     pairing_exists = TabSettings.get("pairing_released", 0) == 1
     printable = printable
+    public_debaters = TabSettings.get("debaters_public", 1)
```
BenMusch, line 417: maybe `debater_team_memberships_public`?

mittab/templates/outrounds/pretty_pairing.html
```diff
@@ -31,8 +31,22 @@
...
+						{{pairing.gov_team.debaters_display}}
+					</div>
+				</td>
+				<td>
+					<div class="teamNames">
+						{{pairing.opp_team.display}}
+					</div>
+					<div class="memberNames" style="display: none;">
+						{{pairing.opp_team.debaters_display}}
+					</div>
```
BenMusch, line 48: If `debaters_public` is false, we shouldn't put this in the HTML at all, even with `display: none`

JoeyRubas:
Good stuff, ran locally and worked great for in-rounds and outrounds. For context for Ben and any other devs who might be off the league, @Speedrace4 had a popular Chrome extension before this with similar functionality on nu-tab, but this seems like a big upgrade to me (and incidentally, will probably be lighter on the servers!)

assets/js/displayMemberNames.js
JoeyRubas: Super elegant idea, didn't know you could approach it that way. Excited to see this!

Speedrace4: I tried this, but for some strange reason it will only update one cell per row and only that cell

```
function changeDisplayNames() {
  const teamNamesCurrentlyDisplayed = document.getElementById("name_display_toggle").checked

  if (teamNamesCurrentlyDisplayed) {
      const teamNames = document.getElementsByClassName("team-name-visible")
      for (let i = 0; i < teamNames.length; i++) {
          const element = teamNames[i];
          element.classList.remove('team-name-visible');
          element.classList.add('team-name-hidden');
      }
      const memberNames = document.getElementsByClassName("member-names-hidden")
      for (let i = 0; i < memberNames.length; i++) {
          const element = memberNames[i];
          element.classList.remove('member-names-hidden');
          element.classList.add('member-names-visible');
      }
  } else {
      const teamNames = document.getElementsByClassName("team-name-hidden")
      for (let i = 0; i < teamNames.length; i++) {
          const element = teamNames[i];
          element.classList.remove('team-name-hidden');
          element.classList.add('team-name-visible');
      }
      const memberNames = document.getElementsByClassName("member-names-visible")
      for (let i = 0; i < memberNames.length; i++) {
          const element = memberNames[i];
          element.classList.add('member-names-hidden');
          element.classList.remove('member-names-visible');
      }
  }
}
```

```
.team-name-visible, .member-names-visible
{
    display: block;
}


.team-name-hidden, .member-names-hidden
{
    display: none;
}
```
(Had to remove the < because the markdown was trying to actually render it)
```
          td>
                div class="team-name-card team-name-visible">
                    {{pairing.gov_team.display}}
                /div>
                {% if public_debaters %}
                div class="member-names-card member-names-hidden">
                    {{pairing.gov_team.debaters_display}}
                /div>
                {% endif %}
            /td>
            td>
                div class="team-name-card team-name-visible">
                    {{pairing.opp_team.display}}
                /div>
                {% if public_debaters %}
                div class="member-names-card member-names-hidden">
                    {{pairing.opp_team.debaters_display}}
                /div>
                {% endif %}
            /td>
```

Speedrace4: I'll take another look tomorrow and see if I can figure it out

Speedrace4: Alternative idea: instead of two divs within each cell, create a new column for member names and team names and toggle the entire column

Speedrace4: I think I'll keep my method; the difference between the two methods is quite small, and this method keeps glitching and not updating all of the classes for reasons I just don't have the free time to figure out. The column idea I proposed doesn't work because they're structured in rows, so I'd still have to loop.

assets/js/displayMemberNames.js
```diff
@@ -0,0 +1,34 @@
...
+      const element = teamNames[i];
+      element.style.display = "none";
+    }
+    const memberNames = document.getElementsByClassName("memberNames");
+    for (let i = 0; i < memberNames.length; i += 1) {
+      const element = memberNames[i];
+      element.style.display = "block";
+    }
+  }
+}
```
JoeyRubas, line 30: I'm not up to spec on best practices for JS, but this worked for me locally

```suggestion
  const teamNamesCurrentlyDisplayed = !document.getElementById("names").checked

  let styleSheet = document.getElementById("dynamicStyles");
  
  styleSheet.innerHTML = teamNamesCurrentlyDisplayed
      ? `.teamNames { display: block !important; } .memberNames { display: none !important; }`
      : `.teamNames { display: none !important; } .memberNames { display: block !important; }`;

```

Then you add this to pairing_display.html and pretty pairing.html
`<style id="dynamicStyles" type="text/css">
    </style>`

assets/js/displayMemberNames.js
```diff
@@ -0,0 +1,34 @@
+import $ from "jquery";
+
+function changeDisplayNames() {
+  const checkBox = document.getElementById("name_display_toggle");
+  const teamNamesCurrentlyDisplayed = !checkBox.checked;
```
JoeyRubas, line 5: I'm not up to spec on best practices for JS, but this worked for me locally. Avoids the looping but no clue if its technically idiomatic.

```suggestion
  const teamNamesCurrentlyDisplayed = !document.getElementById("names").checked

  let styleSheet = document.getElementById("dynamicStyles");
  
  styleSheet.innerHTML = teamNamesCurrentlyDisplayed
      ? `.teamNames { display: block !important; } .memberNames { display: none !important; }`
      : `.teamNames { display: none !important; } .memberNames { display: block !important; }`;

```

Then you add this to pairing_display.html and pretty pairing.html
`<style id="dynamicStyles" type="text/css">
    </style>`

assets/js/displayMemberNames.js
```diff
@@ -0,0 +1,34 @@
...
+    for (let i = 0; i < teamNames.length; i += 1) {
+      const element = teamNames[i];
+      element.style.display = "none";
+    }
+    const memberNames = document.getElementsByClassName("memberNames");
+    for (let i = 0; i < memberNames.length; i += 1) {
+      const element = memberNames[i];
+      element.style.display = "block";
+    }
+  }
```
JoeyRubas, line 29: I'm not up to spec on best practices for JS, but this worked for me locally. Avoids the looping but no clue if its technically idiomatic.

```suggestion
  let styleSheet = document.getElementById("nameStyling");
  
  styleSheet.innerHTML = teamNamesCurrentlyDisplayed
      ? `.teamNames { display: block !important; } .memberNames { display: none !important; }`
      : `.teamNames { display: none !important; } .memberNames { display: block !important; }`;
```

Then you add this to pairing_display.html and pretty_pairing.html
`<style id="nameStyling" type="text/css">
    </style>`

assets/css/pairing-display.scss
```diff
@@ -124,3 +124,9 @@ body
 {
     float: right;
 }
+
+.show-team-names .teamNames { display: block !important; }
```
BenMusch, line 128: nit: `team-names` and `member-names` for consistency

mittab/apps/tab/outround_pairing_views.py
```diff
@@ -566,6 +566,7 @@ def pretty_pair(request, type_of_round=BreakingTeam.VARSITY, printable=False):
 
     sidelock = TabSettings.get("sidelock", 0)
     choice = TabSettings.get("choice", 0)
+    debater_team_memberships_public = TabSettings.get("debaters_public", 1)
```
BenMusch, line 569: Looking into it more, I think we can actually replace this with just the `teams_public` setting **but** the default here should be 0, not 1. Then in `settings.yaml`, update the description to note that it enables the debater names being listed on the pairings page

BenMusch (commented):
LGTM! Some very minor comments and then this is good to merge. Thanks for the work on this!

mittab/apps/tab/outround_pairing_views.py
Speedrace4: I have it this way because the debaters_display function pulls this value, and if it's false then it returns nothing, so we want the same value to make sure there is no button if the debaters aren't public

Speedrace4:
Just squashed all the commits into one to make it cleaner

BenMusch (approved):

PR #356, opened 2025-02-26
Simulate rounds

JoeyRubas:
Minor change but I found it pretty helpful. Probably most nice for development sake, but good for demos/learning tab/running the test tournament to re-familiarize yourself.


BenMusch (commented):
OOC, when are you using this sort of feature? I think the more standard way to do this would be to define something thin the `settings.py` file like:

```
DEVELOPMENT = os.environ.get("MITTAB_ENV") == "development"
```

And then to set that env var locally and have this button show up when it's set. The biggest advantage here is that we don't expose a debugging/testing feature to anyone when the app is actually deployed.

The slight caveat here is that if it's used when people set-up test tournaments and play around, then it may be useful to expose it for the test tournament deployments. In that case, maybe the `TabSetting` is best but we probably wanna still protect it with an env-var guard clause

JoeyRubas:
Brief follow-up before I add this: is "MITTAB_ENV" already something that's being set, or would automatically get set?  I looked around briefly and didn't see it anywhere here or in the deployer.

If not, how about this for a similair outcome:
`if os.environ.get(DEBUG)=="true" or os.environ.get(TOURNAMENT_NAME).endswith('-test')`

Should be a similair effect?

BenMusch:
`MITTAB_ENV` is not currently set, but I'd argue it's a flaw in our infrastructure that it isn't. I'll make that change today

BenMusch:
The fact that test tournaments end in `-test` is an implementation detail of the deployer, so I'd rather the app itself not depend on that

BenMusch:
Should be set on the deployer now: https://github.com/MIT-Tab/mittab-deploy/commit/8b7489748ce742e3768027bc6d99fe44109d5e17

BenMusch (approved):

PR #363, opened 2025-02-28
New Judge Options: remove judge, assign chair

JoeyRubas:
People keep asking for this, and the existing dropdown seemed like the easiest place to put it without interfering too much with the existing UI.

One design note:
This combines outround and inround functionality more than is ususal in the existing code base. As a result the endpoint for assign chair and remove judge for outrounds is in the pairing_views file, the outround JS is in the pairing.js file etc (mostly because the part that handles outounds is just one if statement that switches the URL or object).

Seemed preferable to repeating a lot of code, but probably goes a bit against the grain of the current design pattern, so thought it was worth flagging.

### DIFF
--- a/assets/js/pairing.js
+++ b/assets/js/pairing.js
@@ -168,6 +168,62 @@ function togglePairingRelease(event) {
   });
 }
 
+function handleJudgeRemoveClick(event) {
+  event.preventDefault();
+  const roundId = $(this).attr("round-id");
+  const judgeId = $(this).attr("judge-id");
+
+  const isOutround = $(this).hasClass("outround-judge-remove");
+  const endpointPrefix = isOutround ? "outround" : "round";
+
+  function handleRemoveSuccess(response) {
+    if (response.success) {
+      window.location.reload();
+    } else {
+      alert("Failed to remove judge.");
+    }
+  }
+
+  function handleRemoveError() {
+    alert("Error removing judge.");
+  }
+
+  $.ajax({
+    url: `/${endpointPrefix}/${roundId}/remove_judge/${judgeId}/`,
+    dataType: "json",
+    success: handleRemoveSuccess,
+    error: handleRemoveError
+  });
+}
+
+function handleChairClick(event) {
+  event.preventDefault();
+  const roundId = $(this).attr("round-id");
+  const judgeId = $(this).attr("judge-id");
+
+  const isOutround = $(this).hasClass("outround-judge-chair");
+  const endpointPrefix = isOutround ? "outround" : "round";
+
+  function handleAssignSuccess(response) {
+    if (response.success) {
+      window.location.reload();
+    } else {
+      alert("Failed to assign chair.");
+    }
+  }
+
+  function handleAssignError() {
+    alert("Error assigning chair.");
+  }
+
+  $.ajax({
+    url: `/${endpointPrefix}/${roundId}/assign_chair/${judgeId}/`,
+    dataType: "json",
+    success: handleAssignSuccess,
+    error: handleAssignError
+  });
+}
+
 $(document).ready(() => {
   populateTabCards();
   $("#team_ranking").each((_, element) => {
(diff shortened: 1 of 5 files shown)


assets/js/pairing.js
```diff
@@ -168,6 +168,62 @@ function togglePairingRelease(event) {
...
+  event.preventDefault();
+  const roundId = $(this).attr("round-id");
+  const judgeId = $(this).attr("judge-id");
+
+  const isOutround = $(this).hasClass("outround-judge-remove");
+  const endpointPrefix = isOutround ? "outround" : "round";
+
+  function handleRemoveSuccess(response) {
+    if (response.success) {
+      window.location.reload();
```
BenMusch, line 181: This can be a follow-up PR (or totally abandoned), but two ideas here to make for a nicer UX on the happy path:

1. We can manually update the UI using jQuery to remove the judge from the list
2. We can add a round ID `id` to the DOM and redirect to `currentUrl#id` to auto-scroll to the round that we just removed the judge from


e.g. if a round looked like:

```
<div class="round-display" id="round-123">...round info...</div>
```

And you visit `mittab.com/pairings#round-123`, it will auto-scroll to that div after a refresh. Something worth considering

BenMusch: this is also all, obviously, pointing out how outdated the mit-tab JS stack is. I'm very open to any PRs to incrementally introduce react, vue, or svelte if that's the sort of thing you may enjoy tinkering with

assets/js/pairing.js
```diff
@@ -181,4 +237,16 @@ $(document).ready(() => {
...
+    "click",
+    ".judge-chair, .outround-judge-chair",
+    handleChairClick
+  );
+
+  $(document).on(
+    "click",
+    ".judge-remove, .outround-judge-remove",
+    handleJudgeRemoveClick
+  );
```
BenMusch, line 251: Just for consistency

```js 
 $( ".judge-chair, .outround-judge-chair").click(
    handleChairClick
  );

  $(".judge-remove, .outround-judge-remove").click(
    handleJudgeRemoveClick
  );
```

mittab/apps/tab/pairing_views.py
```diff
@@ -656,3 +656,36 @@ def delete_obj(obj_type):
...
+            round_obj.judges.remove(judge)
+            if round_obj.chair == judge:
+                if round_obj.judges.count() > 0:
+                    round_obj.chair = round_obj.judges.order_by("-rank").first()
+                else:
+                    round_obj.chair = None
+            round_obj.save()
+            return JsonResponse({"success": True})
+        except ValueError:
+            return redirect_and_flash_error(request, "Judge could not be removed")
```
BenMusch, line 676: For the try/catch:

What's the condition that causes this? Is there a way we can check for it in a way that isn't using a `try/catch`?

For the DB usage:

As written, I think this issues a few unneeded DB queries. Specifically: `.judges.count()` does a `select count(*) from judges ...` and `judges.order_by("-rank").first()` does a `select * from judges ... order by ... limit 1`

I think a better pattern would be:

```python
all_judges = round_obj.judges.all().order_by("-rank")
if judge in all_judges:
    round_obj.judges.remove(judge)
    if round_obj.chair == judge:
        if len(all_judges):
            round_obj.chair = all_judges[0]
        else:
            round_obj.char = None
        round_obj.save()
```

BenMusch (commented):
Minor comments -- will do some testing myself this weekend as well

assets/js/pairing.js
JoeyRubas: Makes sense to me. For what its worth, the refresh has been super unintrusive in my testing, (which I'm guessing is mostly a result of lots of smart cacheing logic in chrome). The scroll position on the page for example doesn't seem to change. The only part of the page that visibly refreshing are the tabcard/results summaries under each teams name, which doesn't interfere too much. Probably worth testing in other browser environments before committing though, I'll take a look in edge and firefox if I have a chance tonight.

assets/js/pairing.js
JoeyRubas: I tried it locally, and it doesn't work if I add them to the main `$(document).ready(() =>` block at the bottom of the document. Because the judge dropdown is dynamically loaded, the document binding is fine, but the .click binding needs to be redone every time a new element is added to the page.

It works fine if you add it to the AJAX success, but I think that's a little less clean. Let me know if you'd prefer that or the original.

mittab/apps/tab/pairing_views.py
JoeyRubas: Had to add `all_judges.remove(judge)` because otherwise `all_judges[0]` can sometimes be the judge you're trying to remove, but otherwise commiting this.

BenMusch (approved):
LGTM -- testing today as well

PR #365, opened 2025-03-02
Public Team Rankings

JoeyRubas:
Keep getting asked for this. The exclude rounds logic is a bit ugly but I think think that has more to do with the structure of stats than anything else.

Code isn't unreasonably slow or anything, but I am a tad worried about what might happen if lots of people hit it at once. Might investigate some more aggressive cacheing before merging.


mittab/apps/tab/models.py
```diff
@@ -169,6 +169,7 @@ class Team(models.Model):
     break_preference = models.IntegerField(default=0,
                                            choices=BREAK_PREFERENCE_CHOICES)
     tiebreaker = models.IntegerField(unique=True, null=True, blank=True)
+    ranking_public = models.BooleanField(default=True)
```
BenMusch, line 172: Is the idea here that some teams want to explicitly exclude their ranking from being publicly visible? not fully understanding why this is on the `Team` model rather than _just_ the tabsetting

BenMusch: The reason I ask is that I sorta think that per-team ranking overrides might not be worth supporting:

On a technical level, it's a false sense of privacy. Say only 1 or 2 teams exclude themselves from the ranking, it gives the pretense of anonymity when it can really be inferred from the team ranking (or from tab cards)


It also just does not strike me as the type of anonymity we should care too much about supporting. Scratches and deeply personal info are obviously private, we can support privacy to aid the functioning of the tournament with things like team codes. But post-hoc concern about someone's ranking doesn't feel pertinent, esp. with most tournaments publishing tab cards

JoeyRubas: Yeah this was the intention. It was modeled after a few recent tournaments, which have published team rankings between rounds with opted out teams blacked out like this:
image
Not entirely clear on the motivation, or how different it is than tab cards, but I'm guessing it has something to do with either
a. your opponents can get a good idea of who they're hitting, which you might not want
b. people don't want the public gawking at their results before they've already left the tournament

BenMusch: Got it -- so the idea is more that teams are following this live rather than a post-tournament ranking page?

JoeyRubas: yea exactly. The intended behavior is after all ballots for a round are in, you can see the rankings of teams between rounds,

BenMusch: Got it, this LGTM to me then. Will play around and do a more specific pass at review some time this week

PR #366, opened 2025-03-03
Django 3.2

JoeyRubas:
Minimal first step to begin upgrading to more recent dependencies. Django 2.2 only allows up to Python 3.8, but 3.2 gets us all the way to Python 10 support as a next step. Other packages removed and updated only as strictly necessary.


mittab/apps/tab/migrations/0001_initial.py
```diff
@@ -44,9 +43,6 @@ class Migration(migrations.Migration):
                                   serialize=False,
                                   verbose_name='ID')),
                 ('name', models.CharField(max_length=30, unique=True)),
-                ('phone',
-                 localflavor.us.models.PhoneNumberField(blank=True,
-                                                        max_length=20)),
```
JoeyRubas, line 49: localflavor had some migrations in all versions that support 3.2 that break this code. The model is removed in the 0002 migration, and localflavor is used nowhere else in the repo, so better to remove it than update the removed model.

BenMusch (approved):
LGTM -- testing locally and will deploy a test tournament before merging

PR #376, opened 2025-03-14
HTML Linting

JoeyRubas:
Add linting and formatting for HTML to the CircleCi, and apply to all HTML templates. Should make HTML formatting/development much easier. Jinja/django specific compatible linter used. Most linting problems are fixable with built in formatter `djlint . --reformat --quiet`

Essentially all changes made with the autoformatter, with main exceptions of
1. add closing to orphan tags
2. add language to HTML tags
3. add name to endblock stags (eg. `{% endblock %} -> `{% endblock content %}`

Edit: Cherrypicked away from the python 3.10 branch. No longer depends on any other branches.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,6 +103,9 @@ jobs:
       - run:
           name: Lint -- JS
           command: npm run lint
+      - run:
+          name: Lint -- HTML
+          command: pipenv run djlint .
   docs:
     docker:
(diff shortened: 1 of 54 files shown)


JoeyRubas:
Closing this for now. We should still do this, but the templates have changes so much this will be easier to do in a new branch than try to resolve all these merge conflicts

PR #377, opened 2025-03-14
Remove local() from views

JoeyRubas:
Code quality nit refrenced in the pylint config


JoeyRubas:
Repo has diverged too much. Moving to an issue, and then applying in a new PR later

PR #381, opened 2025-08-06
Apda ids

JoeyRubas:
Small step towards creating an automatic way to upload tournament results onto [standings](http://results.apda.online). Optional setting to add the IDs from standings for debaters and schools into mit-tab. Eventually, I would like to have a publish button which exposes an endpoint to [standings](http://results.apda.online), so that board can upload results with a few clicks


mittab/apps/tab/migrations/0022_auto_20250806_0002.py
```diff
@@ -0,0 +1,23 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('tab', '0021_auto_20231104_1847'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='debater',
+            name='APDA_id',
```
BenMusch, line 15:
```suggestion
            name='apda_id',
```

mittab/apps/tab/migrations/0022_auto_20250806_0002.py
```diff
@@ -0,0 +1,23 @@
...
+            field=models.IntegerField(blank=True, default=-1, null=True),
+        ),
+        migrations.AddField(
+            model_name='school',
+            name='APDA_id',
```
BenMusch, line 20:
```suggestion
            name='apda_id',
```

settings.yaml
```diff
@@ -54,4 +54,9 @@
   name: choice
   description: "A toggle to indicate whether you would like to display who has choice, whether you would like this to be  be indicated on the pairing card, and bolded on the front-facing pairing display"
   value: false
+  type: boolean
```
BenMusch, line 57: any reason to not just always have this on?

JoeyRubas: Idea was just so it's not in the way for unofficial tournament, inhouse tournaments, and I think @roddajohn has mentioned before that some highschool tournaments still use mit-tab as well. Maybe the better approach is to make this an all encompassing "official apda tournament" setting, which would be more descriptive if there's more functionality associated with this.

BenMusch: got it, that makes sense to me! I'm fine with either approach

BenMusch (approved):

PR #386, opened 2025-08-25
Assign judges to outrounds

JoeyRubas:
Posting as draft for visibility because it depends on #385 (both require modifying get sorted pairings to allow outrounds). About 90% tested and functional, but I need to make sure the same judges don't get assigned to varsity and novice rounds happening concurrently.


assets/js/outround.js
```diff
@@ -15,27 +15,33 @@ function cycleChoice(event) {
   });
 }
 
-function populateTabCard(tabCardElement) {
-  const teamId = tabCardElement.attr("team-id");
+function populateTabCard() {
+  const roundNumber = $("#round-number").data("round-number");
+  if (!roundNumber || !$(".outround-tabcard").length) {
+    return;
+  }
```
jordaeday, line 22: should this return an error of some sort?

mittab/libs/assign_judges.py
```diff
@@ -83,6 +83,107 @@ def add_judges():
     Round.judges.through.objects.bulk_create(judge_round_joins)
 
 
+def add_outround_judges(round_type=Outround.VARSITY):
```
jordaeday, line 86: its possible i'm missing it but is there a check to see if there even are enough available judges?

JoeyRubas: Kinda. Before you break to the next round, there's a screen that makes you validate you have enough judges. Hypothetically, you could pass that, and then check out a judge, but in that case this method should fail gracefully and let the user know it was unable to assign judges (this is around line ~147)

PR #387, opened 2025-09-06
Publish results api

JoeyRubas:
Creates an API endpoint to be called by the standings website with all the results data needed to update standings. API isn't authenticated, but this endpoint is only exposed after all the information it has is already public, so this shouldn't be a problem.

### DIFF
--- a/mittab/apps/tab/views.py
+++ b/mittab/apps/tab/views.py
@@ -27,9 +27,11 @@ def index(request):
     debater_list = [(debater.pk, debater.display)
                     for debater in Debater.objects.all()]
     room_list = [(room.pk, room.name) for room in Room.objects.all()]
-    publish_ready = Outround.objects.filter(num_teams=2).exclude(
+    expected_finals = 1+ bool(TabSettings.get("nov_teams_to_break", 4))
+    completed_finals = Outround.objects.filter(num_teams=2).exclude(
         victor=Outround.UNKNOWN
-    ).count() == bool(TabSettings.get("nov_teams_to_break", 4)) + 1
+    ).count()
+    publish_ready = completed_finals == expected_finals
     results_published = TabSettings.get("results_published", False)
 
     number_teams = len(team_list)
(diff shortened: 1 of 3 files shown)


mittab/apps/tab/views.py
```diff
@@ -26,6 +27,10 @@ def index(request):
     debater_list = [(debater.pk, debater.display)
                     for debater in Debater.objects.all()]
     room_list = [(room.pk, room.name) for room in Room.objects.all()]
+    publish_ready = Outround.objects.filter(num_teams=2).exclude(
+        victor=Outround.UNKNOWN
+    ).count() == bool(TabSettings.get("nov_teams_to_break", 4)) + 1
+    results_published = TabSettings.get("results_published", False)
```
jordaeday, line 33: nit: should probably format differently for readability (.exclude() and .count() on separate lines, don't break up the parenthesis)

mittab/apps/tab/views.py
```diff
@@ -455,3 +456,51 @@ def batch_checkin(request):
         "round_numbers": round_numbers,
         "rooms_and_checkins": rooms_and_checkins,
         })
+
+
+def publish_results(request, action):
+    # Convert URL parameter: 0 = unpublish, 1 = publish
```
jordaeday, line 462: nit: should be named in such a way that describes what it does (ie is_published)

mittab/libs/api_standings.py
```diff
@@ -0,0 +1,76 @@
...
+    novice_speakers = novice_speakers[:10]
+
+    return {
+        "varsity_speaker_awards": get_speaker_awards_data(varsity_speakers),
+        "novice_speaker_awards": get_speaker_awards_data(novice_speakers),
+        "varsity_team_placements": get_team_placements_data(Team.VARSITY),
+        "novice_team_placements": get_team_placements_data(Team.NOVICE),
+        "registered_debater_ids": get_registered_debater_ids(),
+        "new_debater_data": get_new_debater_data()
+    }
```
jordaeday, line 76: might be useful for the user to also be able to access a list of all debaters (combining registered and new)

JoeyRubas: Main reason not to was that we're transferring different data for each (new debaters need all their info, because that's being used to add them to the DB, other debaters are already in the DB, so there's no use for that extra data. Shouldn't be a problem b/c the only intended consumer of this endpoint is standings

mittab/apps/tab/views.py
JoeyRubas: Went with something different because I want to make it clear that I'm referring to the state after the endpoint runs, as opposed to the state that it's currently set to, before the endpoint. Should be clearer now nonetheless

PR #392, opened 2025-09-21
Add home buttons

JoeyRubas:
Nice little nit to make the public page feel more connected. Would eventually like a nav bar, but that will require a little more rethinking of the home page and cohesiveness of the public pages, so seems like a project for later

### DIFF
--- a/mittab/templates/ballots/missing_ballots.html
+++ b/mittab/templates/ballots/missing_ballots.html
@@ -4,6 +4,10 @@
 
 {% block content %}
 
+<div class="col-12 mb-3 text-left pl-3">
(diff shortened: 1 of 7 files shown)


mittab/templates/ballots/missing_ballots.html
```diff
@@ -4,6 +4,10 @@
 
 {% block content %}
 
+<div class="col-12 mb-3 text-left pl-3">
+  <a href="/" class="btn btn-sm btn-primary">Back to Home</a>
```
BenMusch, line 8: needs `home-link`?

BenMusch: or does that styling only apply to the footers on the other pages?

JoeyRubas: For whatever reason missing_ballots is using the same CSS as the logged in/tab staff pairing views, not the CSS the rest of the public views use. Since this has a bootstrap import, figured it would be simpler to use the bootstrap classes, they look more or less the same.

BenMusch (approved):

PR #395, opened 2025-09-23
Bracket

JoeyRubas:
nice outrounds bracket. I wanted to use the JS module because it does a LOT of work for us, and is decently well maintained. Unfortunately, it demands a quite complex data schema, hence the giant python modules. Nonetheless, this is probably still more lightweight than doing this 100% on our own. At a minimum, since all the bulky code is completely separate from existing functionality, there's not much risk.

### DIFF
--- a/assets/js/bracket.js
+++ b/assets/js/bracket.js
@@ -1,4 +1,5 @@
 import "../css/bracket.scss";
+import $ from "jquery";
 
 const SELECTOR = "#brackets-viewer-container";
 const TAB_CONFIG = [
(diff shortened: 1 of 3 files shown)


assets/js/bracket.js
```diff
@@ -0,0 +1,152 @@
...
+  const $backdrop = $("#match-modal-backdrop");
+  const $content = $("#modal-content");
+  if (!$modal.length || !$backdrop.length || !$content.length) return;
+
+  const judges =
+    metadata.judges && metadata.judges.length
+      ? metadata.judges
+          .map(judge => {
+            const name = judge.is_chair
+              ? `<strong>${judge.name}</strong>`
```
BenMusch, line 75: This likely isn't a threat we're at risk for, but be warned that this makes us vulnerable to XSS/HTML injenction. Probably work sanitizing

Issue #403, opened 2025-09-30
Add proportional judge assignments


JoeyRubas:
Building on #402, judge assignment should better respect proportional differences. 

For example:
- If our judge pool is composed of have 10 judges ranked 90-99, and 10 judges ranked 1-10, the difference between the best and second best judge should be considered much smaller than the difference between the 10th and 11th best.
- If our top-5 teams are all 1 speak apart, and number 6 is 10 speaks below them, the difference between the first and second should be smaller for pairings sake than the difference between 5 and 6

Issue #274, opened 2019-10-22
Settings page improvements


BenMusch:
Two areas from improvement:

1. Add a `section` grouping, which can either be added like:

```yaml
high_opp_penalty:
  ...
  section: Pairing algorithm
```

Or like:

```yaml
- section: Pairing algorithm
  description: Modify the weights of the pairing algorithm
  settings:
      - high_opp_penalty:
         ...
```

and then use this to group in the UI

2. Add another type: `enum` (like `boolean`)

Based on recent ~~doomthreads~~ constructive discussions it seems like there's strong support to at least have the option to pair judges based on a middle-out style, and some other tweaks like this. Though a boolean may make sense in some instances, an enum is probably a better choice so support the possibility of having more than 1 way to pair judges.

The yaml file could look like:

```yaml
judge_pairing_strategy:
  type: enum
  options:
    - "Middle out"
    - "Top down"
    - "Randomized within bracket"
```

On the UI, this would lead to a select box where the display is the listed option and the value is the index in the list

Additionally, we could add support for a function on the `TabSetting` module like:

```python
def coerced_value(self):
  # parse the yaml config based on self.name
  if yaml_config["type"] == "boolean":
    return self.value == 1
  elif yaml_config["type"] == "enum":
    return yaml_config["options"][self.value]
  else:
    return self.value
```

PR #404, opened 2025-09-30
judge improvements

### LINKED ISSUE #402: Very strange judge assignment behavior at small tournaments
For the past few months, I’ve been hearing complaints about strange judge-pairing behavior at smaller tournaments. I wasn't fully convinced until I encountered some clearly suboptimal judge assignments TOing this last weekend. After running extensive tests and consulting with some other APDA SWEs who were curious, I think I’ve identified the culprit. The bug has likely been around since `calc_weight` was [[last modified](https://github.com/MIT-Tab/mit-tab/commit/7c0d02e888c0afc57041d8253a534dcb22ff95a4)](https://github.com/MIT-Tab/mit-tab/commit/7c0d02e888c0afc57041d8253a534dcb22ff95a4) in 2019.

The problematic line is:

```python
return -1 * abs(judge_i - (-1 * pairing_i))
```

Simplified, this is just:

```python
-1 * (judge_i + pairing_i)
```

Because every potential matching in MWM considers all rounds, the sum of all `pairing_i` values is constant. That means the algorithm is effectively minimizing the sum of `judge_i` values (i.e., the ranks of the judges being assigned). This does not seem like the intended functionality: it's effectively agnostic to where judges are assigned, and only tried to make sure the best judges are somewhere in the pairing. At big tournaments, where each judge could be paired into many potential rounds, this actually works out okay, because MWM then breaks ties by picking the lexicographically lowest arrangement, which often mimics correct behavior. But at small tournaments with many school-affiliation conflicts, the results can diverge badly. Consider this example:

* **Option A**: The best judge (`judge_i = 0`) is placed in the top room. A mid-low judge (`judge_i = 15`) is paired out, instead pairing a nearly identically mid-low ranked (`judge_i = 16`) into bottom room.
* **Option B**: The best judge (`judge_i = 0`) goes to the bottom room. The mid-low judge (`judge_i = 15`) is now paired into the *top* room, while  (`judge_i = 16`) is paired out.

Clearly, Option A is preferable: having the top judge in the top room outweighs a one judge rank difference. But under the current logic, MWM will instead choose Option B.

Lastly: the algorithm doesn’t handle ties in judge ranks gracefully. If you have ten judges all ranked 99, it arbitrarily orders them, then assigns significant weight to those random differences. It would make more sense to give them all the same judge_i

JoeyRubas:
Fixes #402 

- Replace the judge assignment logic with a new default weighing, that prioritizes good judges in good rooms, without replicating the pre-2019 issue (deliberately selecting worse judges for worse rooms)
- Add support to settings for dropdown fields
- Add a dropdown field setting to go back to old judge pairing for backwards compatibility
- Allow changing the re-judge penalty in settings
- Give the same judge_i to judges with tied ranks

TODO: 
- #274 has to be up next, the settings page is unwieldy and very simple low risk settings (show or don't show teams list) should be surfaced before more custom/advanced functionality
- Lower priority, but #403 probably should also get addressed this semester
- Judges should be moved into a utils folder and broken up and modularized more

### DIFF
--- a/mittab/libs/assign_judges.py
+++ b/mittab/libs/assign_judges.py
@@ -7,7 +7,9 @@ class JudgePairingMode:
     DEFAULT = 0
     CLASSIC = 1
 
-def construct_judge_scores(judges):
+def construct_judge_scores(judges, mode=JudgePairingMode.DEFAULT):
+    if mode == JudgePairingMode.CLASSIC:
+        return list(range(len(judges)))
     judge_scores = []
     judge_score = -1
     previous_rank = 0
@@ -21,6 +23,7 @@ def construct_judge_scores(judges):
 
 def add_judges():
     current_round_number = TabSettings.get("cur_round") - 1
+    mode = TabSettings.get("judge_pairing_mode", JudgePairingMode.DEFAULT)
 
     # First clear any existing judge assignments
     Round.judges.through.objects.filter(
@@ -46,7 +49,7 @@ def add_judges():
 
     # Order the judges and pairings by power ranking
     judges = sorted(judges, key=lambda j: j.rank, reverse=True)
-    judge_scores = construct_judge_scores(judges)
+    judge_scores = construct_judge_scores(judges, mode)
     pairings.sort(
         key=lambda x: tab_logic.team_comp(x, current_round_number), reverse=True
     )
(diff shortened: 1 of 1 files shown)


JoeyRubas:
Got this one reviewed informally with a handful of people over messenger, two of which tested on their local machines. GW Inhouse needs it today, so I'm merging

PR #472, opened 2025-10-28
Linters

JoeyRubas:
git diff is deceptively huge. We're 7k lines of package-lock.json changes because we're jumping 3+ major versions on our lint dependencies. Tried to find a workaround but we'll probably just have to stomach it, it doesn't touch prod so it shouldn't be life or death. Actual linting changes are about 400 lines of real code


JoeyRubas:
QA testing looks good. Deployment from local deployer copy looks good. Gonna call that good to go

PR #504, opened 2026-02-15
seperate login level for APDA board

JoeyRubas:
seperate login level for APDA board for results entry. Login level is limited to publishing/unpublishing results, editing APDA ids


JoeyRubas:
@copilot fix failing test

